PR #18, opened 2018-07-03
Interface send eth

hsurf22:


src/components/Development/Elements/Elements.vue
```diff
@@ -0,0 +1,36 @@
+<template>
```
gamalielhere: @hsurf22 @kvhnuke What's this template for?

gamalielhere: @hsurf22 @kvhnuke bump.

gamalielhere:
Closing since the only thing here is an initial setup for the styleguide.

PR #90, opened 2018-09-11
devop/add-more-nodes

gamalielhere:
### Devop  
* [x] Updated CHANGELOG.md  
* [x] Add PR label


src/layouts/InterfaceLayout/components/InterfaceNetworkModal/InterfaceNetworkModal.vue
```diff
@@ -31,11 +31,11 @@
           :key="key + index"
           class="content-block">
           <div class="network-title">
-            <img 
-              v-if="key === 'ROP' || key === 'RIN' || key === 'KOV'" 
+            <img
+              v-if="key === 'ROP' || key === 'RIN' || key === 'KOV' || key === 'ATH'"
```
SteveMieskoski: I think this should be a computed value so, we can use an array or something instead of having them hard coded.

gamalielhere: @SteveMieskoski actually, it can just be a boolean. let me update it.

src/layouts/InterfaceLayout/components/InterfaceNetwork/InterfaceNetwork.vue
```diff
@@ -66,6 +66,11 @@ export default {
   methods: {
     networkModalOpen() {
       this.$children[0].$refs.network.show();
+    },
+    iconExist(name) {
+      if (name === 'ROP' || name === 'RIN' || name === 'KOV' || name === 'ATH')
```
kvhnuke: you should get this by importing network types :), it is better not to hardcode

PR #109, opened 2018-09-25
Testing/initial setup

Royalone-mobile:
components, containers, and layouts unit test-step1


SteveMieskoski:
included in PR #116

PR #149, opened 2018-10-31
Devop/nonce cacher

gamalielhere:
### Devop

- [x] Updated CHANGELOG.md  
- [x] Add PR label

### DIFF
--- a/src/wallets/web3-provider/methods/eth_getTransactionCount.js
+++ b/src/wallets/web3-provider/methods/eth_getTransactionCount.js
@@ -21,7 +21,6 @@ export default async ({ payload, requestManager }, res, next) => {
     Math.round(
       (new Date().getTime() - store.get(utils.sha3(addr)).timestamp) / 1000
     ) / 60; // Get minutes
-  console.log(lastFetch);
   if (lastFetch < 1) {
     fetchedNonce = storedNonce;
   } else {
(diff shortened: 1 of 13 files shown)


src/wallets/web3-provider/methods/eth_getTransactionCount.js
```diff
@@ -0,0 +1,45 @@
...
+    storedNonce = store.get(utils.sha3(addr)).nonce;
+  }
+  const lastFetch =
+    Math.round(
+      (new Date().getTime() - store.get(utils.sha3(addr)).timestamp) / 1000
+    ) / 60; // Get minutes
+  if (lastFetch < 1) {
+    fetchedNonce = storedNonce;
+  } else {
+    fetchedNonce = await ethCalls.getTransactionCount(addr);
```
kvhnuke: You only save the fetchednonce if it is higher than the stored one, right now for example if the last checked is over 1 min, it'll replace the stored nonce with the fetchednonce which can result in transactions with the same nonce

src/wallets/web3-provider/methods/eth_getTransactionCount.js
```diff
@@ -0,0 +1,45 @@
...
+    store.set(utils.sha3(addr), {
+      nonce: Number(fetchedNonce),
+      timestamp: +new Date()
+    });
+  }
+
+  if (storedNonce > Number(fetchedNonce)) {
```
kvhnuke: Let's use bignumber.js instead of `Number()` to prevent overflow

Issue #451, opened 2019-01-14
Hide unsuitable menu options


alepc253:
- **I'm submitting a ...**

  - [x] Feature request
  - [ ] Bug report

- **Feature Request**

  - Some features are not available or suitable for certain networks. So, it would be useful to hide those features by configuration in order to avoid usability problems.

hsurf22:
Fixed

alepc253:
@hsurf22 where can I find this fix?

PR #452, opened 2019-01-14
Hide menu options that are not available for the selected network

alepc253:
### Feature

- [ ] Updated CHANGELOG.md
  - [x] Link to issue: [#451](https://github.com/MyEtherWallet/MyEtherWallet/issues/451)
- [x] Add PR label

### DIFF
--- a/src/layouts/InterfaceLayout/components/InterfaceSideMenu/InterfaceSideMenu.vue
+++ b/src/layouts/InterfaceLayout/components/InterfaceSideMenu/InterfaceSideMenu.vue
@@ -6,7 +6,7 @@
     </div>
     <div class="side-menu">
       <ul>
-        <li v-for="(tab, idx) in tabData" :key="tab.name + idx">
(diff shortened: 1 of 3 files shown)


kvhnuke:
This is great! Thanks. What do you think about doing it the opposite way, for example each network will have supported tabs! since some networks might support some dapps not others
```
ETH:{
         supported: {
                    dapps: ["ens", "ens-subdomain"], 
                    send: ["send-online","send-offline"]
                    ...etc
         }
}
```

alepc253:
I think now we have a better solution, we can set supported menu **and** children menu tabs (and, as before, if you don't set this setting in the network configuration, everything is shown).
About dapps, they are not built as tabs so it should be managed with another setting.

alepc253:
@kvhnuke any news about this?

kvhnuke:
we are integrating this in one of our PRs Thanks!

Issue #464, opened 2019-01-15
Add custom default gasPrice


alepc253:
- **I'm submitting a ...**

  - [x] Feature request
  - [ ] Bug report

- **Feature Request**

  - It is necessary for some non ETH networks to specify default `gasPrice` or even `minimumGasPrice`. 
  - RSK has a `minimumGasPrice` setting that defines the minimum gasPrice that needs to be paid to be mined (otherwise is rejected). This value may differ from `eth_gasPrice` (now is the same). 
  - RSK `minimumGasPrice` is not ETH low price (`eth_gasPrice / 2`) so it could result in users choosing  low price option that won't be able to send tx.

alepc253:
There is an open PR about this

PR #465, opened 2019-01-15
Custom regular and minimum gas price

alepc253:
### Feature

- [ ] Updated CHANGELOG.md
- [ ] Is this a user submitted feature request?
  - [x] Link to issue: [#464](https://github.com/MyEtherWallet/MyEtherWallet/issues/464)
- [x] Add PR label


alepc253:
Not that much... If the network doesn't set a minimum gasprice then it uses
gasPrice / 2.
It shouldn't change anything to other networks but RSK

alepc253:
RSK won't work with gasPrice / 2

alepc253:
@kvhnuke could you rethink this? Let me know if you need more info about RSK `minimumGasPrice`

alepc253:
@kvhnuke bad news! this is really necessary to avoid unmined tx because of using "safe low" gas price that will be for sure below `minimumGasPrice`

src/store/actions.js
```diff
@@ -95,6 +95,10 @@ const setGasPrice = function({ commit }, gasPrice) {
   commit('SET_GAS_PRICE', gasPrice);
 };
 
+const setMinimumGasPrice = function({ commit }, minimumGasPrice) {
+  commit('SET_MINIMUM_GAS_PRICE', minimumGasPrice);
+};
+
```
SteveMieskoski: If the minimum gas price is located on the network object then there is not reason to also add a new property to the store/state.

src/layouts/InterfaceLayout/components/InterfaceNetworkModal/InterfaceNetworkModal.vue
```diff
@@ -394,6 +394,15 @@ export default {
...
           this.selectedeNtworkName = network.name;
+          this.selectedNetwork = network;
+          this.$store.dispatch(
+            'setGasPrice',
+            Number(network.type.regularGasPrice || 41)
+          );
+          this.$store.dispatch(
+            'setMinimumGasPrice',
+            Number(network.type.minimumGasPrice)
+          );
```
SteveMieskoski: Let's try not to add a new property (setMinimumGasPrice) to the store/state. (see comment in actions.js)

src/store/mutations.js
```diff
@@ -71,6 +71,11 @@ const SET_GAS_PRICE = function(state, val) {
   store.set('gasPrice', val);
 };
 
+const SET_MINIMUM_GAS_PRICE = function(state, val) {
+  state.minimumGasPrice = val;
+  store.set('minimumGasPrice', val);
+};
+
```
SteveMieskoski: Let's try not to add a new property to the store/state. (see comment in actions.js)

src/store/state.js
```diff
@@ -9,9 +9,14 @@ const network =
...
+const regularGasPrice = network.type.regularGasPrice;
 const gasPrice =
-  store.get('gasPrice') !== undefined ? store.get('gasPrice') : 10;
-
+  store.get('gasPrice') !== undefined
+    ? store.get('gasPrice')
+    : regularGasPrice !== undefined
+    ? regularGasPrice
+    : 41;
+const minimumGasPrice = network.type.minimumGasPrice || 0;
```
SteveMieskoski: Let's try not to add a new property to the store/state. (see comment in actions.js)

alepc253:
@SteveMieskoski let me know if it's ok now, thanks!

alepc253:
@SteveMieskoski anything else that should be addressed?

kvhnuke:
I think this is something RSK clients should handle, as throw an error if the gasPrice is below certain amount. We are trying to keep things universal and if we implement this, we have to implement it to all the other chains as well, since I personally dont like having custom options on certain networks but not others. If you can figure out the minimum gas prices for all the networks we support I can take another look at it.

kvhnuke:
on the second thought, what if we change the current way of calculating the low, medium, high to
low = gasPrice
medium = gasPrice*1.5
high = gasPrice*2

what are your thoughts on that?

alepc253:
@kvhnuke thanks for your comments. About your first comment, we can't know that for every other chain but we can use the same magic number that is used if gasPrice query fails.
About your second approach, it's not the ideal scenario but is much better than the actual one. If you feel confortable with this solution, we can do this now and maybe iterate on it later if a better option comes up.

alepc253:
@kvhnuke I went for the second approach, let me know if this is ok now

kvhnuke:
Thanks @alepc253

src/layouts/InterfaceLayout/components/InterfaceNetworkModal/InterfaceNetworkModal.vue
```diff
@@ -393,6 +393,10 @@ export default {
       this.$store.dispatch('switchNetwork', network).then(() => {
         this.$store.dispatch('setWeb3Instance').then(() => {
           this.selectedeNtworkName = network.name;
+          this.$store.dispatch(
+            'setGasPrice',
+            Number(network.type.gasPrice || 41)
```
kvhnuke: I dont think this is necessary!

PR #507, opened 2019-01-24
Add an integration guide to assist dapp developers

SteveMieskoski:
Add an integration guide to assist dapp developers


gamalielhere:
I would suggest adding a line about keeping the scope of the dapp within the dapps folder i.e "Don't touch anything else unless necessary". Otherwise, this looks good.

PR #537, opened 2019-01-31
Devop/swap revisions 3

SteveMieskoski:
### Devop

- [ ] Updated CHANGELOG.md
- [ ] Add PR label

Improve user experience during recalculation, and display all providers at all times


src/helpers/notificationFormatters/formatters.js
```diff
@@ -38,6 +40,15 @@ const extractErrorMessage = errObj => {
 
 // eslint-disable-next-line no-unused-vars
```
kvhnuke: why is it here?

src/components/Notification/Notification.vue
```diff
@@ -140,7 +144,7 @@ export default {
       account: 'account'
     }),
     sortedNotifications() {
-      this.countUnread();
+      // this.countUnread();
```
kvhnuke: this requires a removal

src/components/Notification/Notification.vue
```diff
@@ -140,7 +144,7 @@ export default {
...
       if (!this.notifications[this.account.address]) return [];
       // eslint-disable-next-line
```
kvhnuke: this too

PR #545, opened 2019-02-01
DApp/ Schedule Transactions

Bagaric:
# DApp: Schedule transactions

This DApp uses the [Ethereum Alarm Clock](https://github.com/ethereum-alarm-clock/ethereum-alarm-clock) protocol to schedule transactions for the future.

## Screenshots
ezgif com-video-to-gif 8


Bagaric:
@kvhnuke Ready to be reviewed!

kvhnuke:
@Bagaric this is awesome! 
can you make the to address support `ens`, you can find the way we do it `sendCurrencyContainer`

kvhnuke:
can you also add a disabled icon
image

Bagaric:
@kvhnuke Addressed your review comments! Should be ready to be merged.

Bagaric:
We've also updated the icon.
screenshot 2019-02-25 at 13 56 55

kvhnuke:
also what do you guys think about moving confirmation hashes, and pending token transfer to a new section, people might miss it in the bottom?

Bagaric:
@kvhnuke Any suggestions on where we could move them?
Toasts are a bit too small. I guess we could use modals, but then they interfere with the automatically-triggered modals when we call `web3.eth.sendTransaction`.

kvhnuke:
I was thinking of going to a new page, like after sending the tx it'll goto the success page

Bagaric:
> I was thinking of going to a new page, like after sending the tx it'll goto the success page

That could be good, but it would break the workflow in case a person wants to schedule multiple similar transactions - as it would clear all the fields upon a new load of the schedule page.

kosecki123:
> > I was thinking of going to a new page, like after sending the tx it'll goto the success page
> 
> That could be good, but it would break the workflow in case a user wants to schedule multiple similar transactions (think scheduling salary transfers to multiple employees at the same time) - as it would clear all the fields upon a new load of the schedule page.

Success page with clear information is more important than possibility of easier rescheduling. 
Dedicated Reschedule button/link could be added later.

Bagaric:
@kvhnuke @kosecki123 I've made the update to show the information in a new screen. Also updated the screenshots sections of this PR - check it out.

src/dapps/ScheduleTransaction/containers/ScheduleView.vue
```diff
@@ -0,0 +1,861 @@
...
+          type: 'number'
+        };
+      }
+    };
+  },
+  computed: {
+    ...mapGetters([
+      'web3',
+      'network',
+      'wallet',
```
kvhnuke: can you remove the `wallet`import

Bagaric: Done!

src/dapps/ScheduleTransaction/containers/ScheduleView.vue
```diff
@@ -0,0 +1,854 @@
...
+      supportedModes: EAC_SCHEDULING_CONFIG.SUPPORTED_MODES,
+      selectedMode: EAC_SCHEDULING_CONFIG.SUPPORTED_MODES[0],
+      deposit:
+        EAC_SCHEDULING_CONFIG.TIME_BOUNTY_DEFAULTS[0] *
+        EAC_SCHEDULING_CONFIG.BOUNTY_TO_DEPOSIT_MULTIPLIER,
+      ethPrice: new BigNumber(0),
+      selectedTimeZone: moment.tz.guess(),
+      selectedCurrency: '',
+      showTokenTransferNotification: true,
+      amountInputOptions() {
```
gamalielhere: These functions that returns the options might be better suited in the `computed` property so whenever the `this` instances inside updates, it also updates instead of being called only once.

PR #557, opened 2019-02-04
Update SUB contract address in tokens-eth.json

substratumservices:
We (Substratum) deployed a new contract and did a token swap. The old contract addresses are frozen and cause wasted gas consumption when transactions fail. This PR updates SUB to the new contract address. If you need additional authentication we can provide a signed message from the owning address of the old or new contract.


gamalielhere:
@substratumservices Please make the PR on [here](https://github.com/MyEtherWallet/ethereum-lists/)

PR #617, opened 2019-02-12
Feature/error toast [5.0.2]

gamalielhere:
### Feature

- [x] Updated CHANGELOG.md
- [x] Add PR label


src/containers/ConfirmationContainer/ConfirmationContainer.vue
```diff
@@ -244,7 +244,7 @@ export default {
           ]);
         })
         .catch(err => {
-          ErrorHandler(err, true);
+          ErrorHandler(err, 3, this);
```
kvhnuke: i dont think you need `this` try importing `import Toasted from 'vue-toasted';` inside the `ErrorHandler` it should work

gamalielhere: I'm not sure how that would work since Toasted is imported as a Vue installer. That's actually why I pass `this` back to the function so I can reference the vue instance outside of the component.

PR #683, opened 2019-02-19
Bug/hd path regex correction

SteveMieskoski:
### Bug
- [ ] Updated CHANGELOG.md
- [x] Add PR label

revert changes/correct hd path parsing


kvhnuke:
added in #680

PR #709, opened 2019-02-22
Devop/update keepkey

### LINKED ISSUE #707: Non-webusb KeepKeys aren't going to work with MEW
We just noticed that keepkey.js was offering to pair with non-webusb KeepKeys, which isn't going to work:

```
An attempt to claim a USB device interface has been blocked because it implements a protected interface class.
```

 That probably spills over to some user frustration... anyway, the fix should be to update to [v1.2.1](https://github.com/keepkey/keepkey.js/releases/tag/v1.2.1), which includes this commit: https://github.com/keepkey/keepkey.js/commit/0c5b570d3c927aea172a6eaa347e2bec49e210a6

gamalielhere:
### Devop

- [x] Updated CHANGELOG.md
- [x] Add PR label

Fixes #707


package.json
```diff
@@ -80,7 +80,7 @@
     "jest-skipped-reporter": "0.0.4",
     "jspdf": "1.5.3",
     "keccak": "2.0.0",
-    "multicoin-address-validator": "git+https://github.com/MyEtherWallet/multicoin-address-validator.git",
+    "multicoin-address-validator": "0.3.9",
```
kvhnuke: switch this back to the github repo

src/layouts/AccessWalletLayout/components/HardwareModal/HardwareModal.vue
```diff
@@ -130,14 +134,34 @@ export default {
       this.items.forEach(item => {
         const u2fhw = ['secalot', 'ledger', 'bitbox'];
         const inMobile = ['secalot', 'keepkey'];
+        const webUsb = ['keepkey'];
+
+        if (webUsb.includes(item.name)) {
+          const disable =
```
keepkeyjon: Alternatively:

```suggestion
          const disable = !window || !window.navigator || !window.navigator.usb;
```

so it starts working if/when other browsers implement it. Not sure if there's a similar trick for detecting u2f support.

gamalielhere: Makes sense. I'm integrating this and adding the `https` check to it.

src/layouts/AccessWalletLayout/components/HardwareModal/HardwareModal.vue
```diff
@@ -130,14 +134,32 @@ export default {
...
```
keepkeyjon: I think you meant:

```
          const disable =
            window.location.protocol !== 'https:' ||
            (!window || !window.navigator || !window.navigator.usb);
```

I.e

1. webusb doesn't work without https
1. webusb doesn't work without window.navigator.usb, which doesn't work if there's no window.navigator, which doesn't work if there's no window.

keepkeyjon: @gamalielhere sorry I missed this by a couple of minutes

gamalielhere: Thanks for the correction! Included the fix on #711

Issue #807, opened 2019-03-14
Currency symbol in type configuration


alepc253:
- **I'm submitting a ...**

  - [x] Feature request
  - [ ] Bug report

- **Feature Request**

  - Actually, in `network\types` you can set `name` and `name_long` for a network. This two configurations are used in different views and, in general, `name` is used as currency symbol.
  - Not necessarily the network name is the currency symbol (for example, `RSK` is the network name and `RBTC` is the currency symbol or ticker).
  - It would be helpful to have a new setting `symbol`, maybe optional, to be shown next to balances, fees, etc.

PR #809, opened 2019-03-14
Add currency symbol to configuration, keep using network name if not defined

alepc253:
### Feature

- [ ] Updated CHANGELOG.md
  - [x] Link to issue: #807 
- [x] Add PR label


kvhnuke:
Lets create a symbol field in all network types and use that, I dont like doing this `const symbol = this.network.type.currencyName || this.network.type.name;`

it should be just `this.network.type.symbol`

alepc253:
@kvhnuke and that `symbol` field should have set the same content as `name` by default, right?

alepc253:
@kvhnuke I addressed your reviews, let me know if that works, thanks!

kvhnuke:
@alepc253 Thank you, this looks good to me, @gamalielhere can you do a final check and see whether we replaced all the instances?

Issue #832, opened 2019-03-20
Show token transfer data (recipient) in notification


alepc253:
- **I'm submitting a ...**

  - [x] Feature request
  - [ ] Bug report

- **Feature Request**

  - It would be helpful that notifications related to token transfers show their info in a user-friendly way.

This is related to issue #821

PR #840, opened 2019-03-21
Show token transfer data in notification body

alepc253:
### Feature

- [ ] Updated CHANGELOG.md
  - [x] Link to issue: #832
- [x] Add PR label

With this changes, it will look like this:

Captura de pantalla 2019-03-25 a la(s) 14 14 10
Captura de pantalla 2019-03-25 a la(s) 14 14 50


gamalielhere:
@alepc253 should probably also change the currency on the Transaction fees section

alepc253:
@gamalielhere I don't think so because gas is still paid in ETH when you transfer tokens (in this particular case, it's true for RSK but that change is in another [PR](https://github.com/MyEtherWallet/MyEtherWallet/pull/824))

alepc253:
@gamalielhere are you ok with this?

PR #898, opened 2019-04-03
Ui/add monospace font

hsurf22:
### Devop

- [x] Updated CHANGELOG.md
- [x] Add PR label
  - Add monospace font


src/assets/google-fonts.css
```diff
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css?family=Inconsolata:400,700');
```
kvhnuke: you have to download this locally and add it to font, we cant reference online fonts

Issue #953, opened 2019-04-15
Initial derivation path should correspond selected network


alepc253:
- **I'm submitting a ...**

  - [x] Feature request
  - [ ] Bug report

- **Feature Request**

  - When accessing by mnemonic phrase or hardware wallet, it would be really user friendly to have selected as initial derivation path one that corresponds with the selected network. 
  - Now, you always get the Ethereum dpath `m/44'/60'/0'/0` whether you selected Ethereum or not.

PR #954, opened 2019-04-15
Initial dpath should correspond selected network

alepc253:
### Feature

- [ ] Updated CHANGELOG.md
  - [x] Link to issue: #953 
- [x] Add PR label


kvhnuke:
instead of making it part of network types, I think this should be a part of bip44 wallet paths, as certain wallets have different paths as the default for certain networks

alepc253:
@kvhnuke any ideas about the implementation? I understand your concern but I don't see how to solve that situation. From the point of view of an end user, it's much better to have selected, at least, one of the possible different paths than always the ethereum path `m/44'/60'/0'/0`.

kvhnuke:
PR #910 slove this issue for ledger, I think the best way to handle this globally is by adding a network field to all the paths in https://github.com/MyEtherWallet/MyEtherWallet/blob/master/src/wallets/bip44/paths.js

then we can loop through the paths and see whether any of the paths matches the current network and choose that as the default?

@alepc253 what do you think?

PR #959, opened 2019-04-16
support 32bits chainId

hackmod:
support 32bits chainId for Ledger hardware wallet.

without this PR, all networks with chainId > 255 do not work

- ~~[x] add all networks supported by Ledger (accidentally removed by 7ec9cd1dda41)~~

Please see https://github.com/kvhnuke/etherwallet/pull/1979

### DIFF
--- a/src/wallets/hardware/ledger/index.js
+++ b/src/wallets/hardware/ledger/index.js
@@ -61,7 +61,7 @@ class ledgerWallet {
     const txSigner = async tx => {
       tx = new ethTx(tx);
       const networkId = tx._chainId;
-      tx.raw[6] = Buffer.from([networkId]);
+      tx.raw[6] = networkId;
       tx.raw[7] = Buffer.from([]);
       tx.raw[8] = Buffer.from([]);
       const tokenInfo = byContractAddress('0x' + tx.to.toString('hex'));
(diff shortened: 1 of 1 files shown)


kvhnuke:
we will be handling the ledger paths differently, please see #910

kvhnuke:
is the 32bit support only needed for ledger or all the wallets? @hackmod

hackmod:
> is the 32bit support only needed for ledger or all the wallets? @hackmod

Current Ledger firmware returns only 1-byte truncated signature `v`. so we can't use `v` byte directly for chainId > 255 cases. (as described at PR https://github.com/kvhnuke/etherwallet/pull/1979)

- Trezor: almost same bug already fixed at the firmware level. but we need to fix some additional stuff to support full rang 32bits chain_Id (see also https://github.com/kvhnuke/etherwallet/pull/2011 https://github.com/trezor/trezor-mcu/pull/399)
- other HD wallets: I'm not sure but I think there's a similar bug could exist.

hackmod:
rebased, drop needless fix.

hackmod:
without this fix, the following chains with `chainId` > `255` are not working with the Ledger nano S.

Chain | `ChainID`
-------|---------
CLO | 820
ATH | 1620
EGEM | 1987
REOSC | 2894
WEB | 24484
ESN | 31102
SUM | 72106
AKA | 200625
ETHO | 1313114
MUSIC | 7762959
IOLITE | 18289463
PIRL | 3125659152

PR #976, opened 2019-04-19
Bug/web3 wallet support

gamalielhere:
### Bug

- [x] Updated CHANGELOG.md
- [x] Add PR label


src/containers/FooterContainer/FooterContainer.vue
```diff
@@ -220,7 +220,8 @@ export default {
             },
             {
               text: 'KeepKey',
-              href: 'http://keepkey.go2cloud.org/aff_c?offer_id=1&aff_id=5561'
+              href:
+                'http://keepkey.myshopify.com?afmc=pi&utm_campaign=pi&utm_source=leaddyno&utm_medium=affiliate'
```
kvhnuke: we also have to edit this page,
https://www.myetherwallet.com/hardware-wallet-affiliates

can we make a js file where we can reference all these urls. That way we just have to import it and only change in one place

PR #1014, opened 2019-04-29
private key improvement

SteveMieskoski:
### Devop

- [x] Updated CHANGELOG.md


src/layouts/AccessWalletLayout/components/PrivateKeyModal/PrivateKeyModal.vue
```diff
@@ -54,7 +52,20 @@ export default {
   computed: {
     ...mapGetters({
       path: 'path'
-    })
+    }),
+    notValid() {
```
kvhnuke: lets reduce it to
```
const _priv = this.privateKey.replace('0x','')
return !isHexString('0x'+_priv, 64)
```

PR #1016, opened 2019-04-29
Bug/address whitespace

gamalielhere:
### Bug

- [x] Updated CHANGELOG.md
- [x] Add PR label


src/helpers/misc.js
```diff
@@ -172,6 +172,10 @@ const isDarklisted = addr => {
   return errObject;
 };
 
+const stripAddr = addr => {
```
kvhnuke: why arent we using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim

gamalielhere: I did not know about it.. My bad

PR #1037, opened 2019-05-02
devop/ledger-custom-paths

gamalielhere:
### Devop

- [x] Updated CHANGELOG.md
- [x] Add PR label


kvhnuke:
@gamalielhere can you remove the `ethereum (Trezor)` path from ethereum app and add it as a custom path which will automatically be loaded

PR #1103, opened 2019-05-23
Fix ENS migrate owner check

kvhnuke:
It is currently checking in the ENS for ownership, however correct method is to check the deed.owner

### Bug

- [x] Updated CHANGELOG.md
- [x] Add PR label


src/dapps/ManageENS/ManageENS.vue
```diff
@@ -354,7 +354,7 @@ export default {
       if (this.parsedTld !== '' && isSupported === undefined) {
         Toast.responseHandler(
           `Domain TLD ${this.parsedTld} is not supported in this node!`,
-          Toast.ERROR
+          Toast.ERROR``
```
gamalielhere: I don't think these tildes are necessary

gamalielhere: I don't think these tildes are necessary

kvhnuke: you are right:) definitely not necessary

PR #1107, opened 2019-05-24
Feature/maker fix

SteveMieskoski:
Revised maker implementation avoiding issues recently discovered


SteveMieskoski:
modifications moved to feature/maker

PR #1209, opened 2019-06-27
BitBox bugfixes: U2F communication; displaying error messages

douglasbakkum:
### Bug

- [x] Updated CHANGELOG.md

---

This PR, in combination with a firmware update https://github.com/digitalbitbox/mcu/pull/283 fixes the BitBox communication error that started with Chrome 73 (which required also changes in Chrome that have since occurred).

We will release a new signed firmware in the next days that can be used to test this PR. If it helps, a live demo of this PR that works with BitBox hardware is also available at https://digitalbitbox.com/mew

This PR also fixes displaying correct error messages (in the red pop up boxes).

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### Bug
+
+- Fix the BitBox hardware wallet communication error introduced with Chrome 73 [#1209](https://github.com/MyEtherWallet/MyEtherWallet/pull/1209)
+- Fix error message display for the BitBox hardware wallet [#1209](https://github.com/MyEtherWallet/MyEtherWallet/pull/1209)
+
 ### Release v5.0.14-hotfix.2
 
 ### Bug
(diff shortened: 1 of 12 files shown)


SteveMieskoski:
@douglasbakkum  we updated the firmware to 7 using the pre-release version of the bitbox-wallet-app and have tried to use the site you are hosting to check (as a work around to the localhost restriction) and we are getting the new error message "MyEtherWallet does not yet support this version of the firmware".  As far as our internal testing, should we use some other version of the firmware to test via localhost.  If so, please shoot us an email with that version and we will get to it. Thanks!

douglasbakkum:
@SteveMieskoski Oh. Sorry. I did not deploy the last commit which bumped the supported firmware version. This is now done and tested. Be sure to clear the browser cache before trying again.

Also note that "Network Error" messages will pop up. This is due to trying to connect to myetherwallet.com by default for block information from a non-myetherwallet origin. So will need to change the network to connect to during the steps to select an address.

PR #1228, opened 2019-07-03
Improvements to Dockerfile

aldoborrero:
### Feature

This PR tries to improve the original Dockerfile in order to have a contanerized version of MEW. As you can see in the Dockerfile, it compiles the code using the previous code but also copy and adds nginx to serve the compiled code.

I left on purpose the `TARGET` argument as it can be used in combination with [modes and environment variables that vue supports](https://cli.vuejs.org/guide/mode-and-env.html), allowing the user to customize different flavors.

Let's keep the discussion open to improve this PR and be useful as much as possible (so we can publish it to Docker hub).

Note: I saw that there are a couple of interesting env variables to customize different parts of the application, so we can update this Dockerfile with all of them and with sane defaults.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,21 @@
-FROM node:8.16.0-jessie
+FROM node:8.16.0-jessie AS build
+
+ARG TARGET=development
 
 ENV NODE_OPTIONS --max-old-space-size=4096
-RUN npm install npm@6.9 -g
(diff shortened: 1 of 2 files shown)


kvhnuke:
you cant do this to us! This Dockerfile is meant to do travis builds. Not to run a MEW locally haahhha

PR #1251, opened 2019-07-09
Additional makerCDP related updates

SteveMieskoski:
### Devop

- [x] Updated CHANGELOG.md
- [x] Add PR label

- Add note to inform users of the destination when moving a cdp.  (shows them the proxy address so, they can check if, for instance, they are using a keepkey)
- Add a check to the swap widget to correct in cases where the to value gets over written after the modal loads.


kvhnuke:
please complete the checklist and PR labels

Issue #1256, opened 2019-07-09
Integrate Lightstreams network support


ggarri:
## Description
Support **Lightstreams Network**, PoA Ethereum compatible network for fast and secure file sharing and responsive DApps.

Homepage: https://lightstreams.network

**Mainnet Network**
Block Explorer: https://explorer.lightstreams.io
RPC Endpoint: https://node.mainnet.lightstreams.io
ChainId: 163

PR #1257, opened 2019-07-09
#1256 Adding Lightstreams Network support

ggarri:
## Description

Support **Lightstreams Network**, Proof of Authority Ethereum compatible network for fast and secure file sharing and responsive DApps.

Homepage: https://lightstreams.network
Issue: #1256 

**Mainnet Network**
Block Explorer: https://explorer.lightstreams.io
RPC Endpoint: https://node.mainnet.lightstreams.io
ChainId: 163


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### Release v5.0.15
 
 ### Bug
+
```
kvhnuke: can you please update the changelog?

src/wallets/hardware/ledger/appPaths.js
```diff
@@ -158,6 +160,11 @@ const appList = [
     network: TT,
     prefixes: ["m/44'/1001'"],
     paths: [thundercore]
+  },
+  {
+    network: PHT,
+    prefixes: ["m/44'/60'"],
+    paths: [lightstreamsNetwork]
```
kvhnuke: Also,  can you add the path to other wallets, such as `Trezor, keepkey..etc`

CHANGELOG.md
ggarri: Done

src/wallets/hardware/ledger/appPaths.js
ggarri: Addressed at https://github.com/MyEtherWallet/MyEtherWallet/pull/1257/commits/dc31bedd0827b198d9dd64d12f6c9c44401ee541

kvhnuke:
@ggarri thanks for the changes however, I dont think you have to add seperate paths for `keepkeyLightstream` and `ledgerLightStream` the reason why we have different paths for `Ethereum` for those is because when they initially released it, different hw wallets started using different paths but in your case you should just add `"m/44'/60'/0'/0"` `lightstreamsNetwork` path to everywhere

ggarri:
Hi @kvhnuke I see where it was my confusion. As our network is fully ethereum compatible I was trying to use exact same paths for our network. I just pushed a new commit which updates the paths for Ledger and Keepkey to use "m/44'/60'/0'/0". 

Let me know if there is something else to do I will try to address it ASAP. Thanks :)

PR #1275, opened 2019-07-19
devop/mewtopia

gamalielhere:
### Devop

- [x] Updated CHANGELOG.md
- [x] Add PR label


src/containers/FooterContainer/FooterContainer.vue
```diff
@@ -221,6 +221,10 @@ export default {
               text: this.$t('common.faqs'),
               to: '/#faqs'
             },
+            {
+              text: 'MEWTopia',
```
kvhnuke: it should be `MEWtopia`

PR #1296, opened 2019-07-30
Renaming "Digital bitbox" to "BitBox"

jstrnbrg:
We renamed our Hardware Wallet line a while ago to just "BitBox" (both B capitalised) and dropped the "Digital" at the front. 
Would be great if you could merge this PR so that the naming is consistent. 

I hope I only made changes effecting the front-end,  ignoring all backend logic where you use the old name. 

In the footer/<language>.json files, it might make sense to drop the "digital" key and only use the "bitbox" key ?


jstrnbrg:
@gamalielhere @kvhnuke @SteveMieskoski when can we expect this to be merged into the production branch? 
Thank you

gamalielhere:
@jstrnbrg this has been merged to production awhile ago. It's already on the live site: https://www.myetherwallet.com/

Issue #1349, opened 2019-08-13
Fix invalid Lightstreams Network Ledger path


ggarri:
- **I'm submitting a ...**

  - [ ] Feature request
  - [x] Bug report

- **Bug Report**
There was an update done by other developer which affected the ledger path for Lightstreams network as we can see [in this commit](https://github.com/MyEtherWallet/MyEtherWallet/commit/8cfd705d594eda9663e4f4b586a1125d70848348#diff-68533bdf45afb84f4330fbb06949ceddL167). This issue intends to update Lightstreams network ledger path to use `m/44'/60'/0'`

EnchanterIO:
Hey guys @kvhnuke / @gamalielhere, Gabriel's colleague here. Could we please get this one merged to fix the current bug for MeW Lightstreams users? 

Let us know if we can help anyhow more or you need more clarification!

PR #1350, opened 2019-08-13
#1349 Restore valid Lightstreams network ledger path

ggarri:
### Bug

  - [x] Link to issue: https://github.com/MyEtherWallet/MyEtherWallet/issues/1349


mikesmo:
Hey @kvhnuke / @gamalielhere. Could we please get this one merged? There is a bug in MEW and no one can use Lightstreams. 

The original code worked but I understand there were some changes to this that introduced the bug.

Let us know if we can help anyhow more or you need more clarification!

PR #1384, opened 2019-08-21
bug/match metamask network on wallet unlock

SteveMieskoski:
### Bug
- [x] Updated CHANGELOG.md
- [X] Add PR label


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ### Release v5.1.2
```
gamalielhere: should probably remove this

src/networks/types/GOERLI.js
```diff
@@ -4,7 +4,7 @@ import { GOERLI } from '../tlds';
 import goerli from '@/assets/images/icons/network.svg';
 
 export default {
-  name: 'GöETH',
+  name: 'GOERLI',
```
kvhnuke: why did you change this to `GOERLI` ?

SteveMieskoski: There was an issue with matching the name as it is different than the type file name.  I don't remember where it was occurring, so I just reverted it.

PR #1394, opened 2019-08-28
Devop/sentry fixes

SteveMieskoski:
### Devop
- [x] Updated CHANGELOG.md
- [x] Add PR label


src/store/actions.js
```diff
@@ -90,7 +90,9 @@ const createAndSignTx = function({ commit }, val) {
 };
 
 const decryptWallet = function({ commit, dispatch }, params) {
-  if (params[0]) {
+  // if the wallet param (param[0]) is undefined or null then all the subsequent setup steps will also fail.
+  // just explicitly stop it here.
+  if (params[0] !== undefined && params[0] !== null) {
```
kvhnuke: can we figure out the point where it sends an empty param[0]

SteveMieskoski: I couldn't reproduce the scenario causing this is practice.  I was able to reproduce the errors that led me to identify this as the source, and then only via explicit assignment.

Issue #1395, opened 2019-08-28
Footer breaks when user expend token list.


hsurf22:
- **Bug Report**
  - What is the current behavior?
   - Footer breaks when user expend token list.

![Screenshot from 2019-08-28 13-13-51](https://user-images.githubusercontent.com/9893774/63890101-233dad00-c997-11e9-9a92-bccf3359845e.png)

PR #1396, opened 2019-08-28
Fix the bug when user expend the token list, footer breaks.

hsurf22:
### Bug

- [x] Updated CHANGELOG.md
  - [x] Link to issue: https://github.com/MyEtherWallet/MyEtherWallet/issues/1395
- [x] Add PR label
   - Fix the bug when user expend the token list, footer breaks.

Screenshot from 2019-08-28 13-13-51


hsurf22:
https://github.com/MyEtherWallet/MyEtherWallet/issues/1395

PR #1409, opened 2019-09-05
Bug/ens

gesseekur:
### Bug

- [x] Updated CHANGELOG.md
- [X] Add PR label


src/dapps/BuySubdomain/BuySubdomain.vue
```diff
@@ -119,7 +122,12 @@ export default {
   },
   methods: {
     debounceInput: web3.utils._.debounce(function(e) {
-      this.domainName = normalise(e.target.value);
+      if (e.target.value.indexOf(' ') >= 0) {
```
kvhnuke: it cannot also have `.` and other invalid characters also `normalise` can fail, may be put normalise inside try catch?

src/dapps/ManageENS/ABI/registrarAbi.js
```diff
@@ -23,30 +23,6 @@ export default [
     payable: false,
```
kvhnuke: I think we can safely delete this whole file? are we referencing this anywhere?

src/dapps/ManageENS/ManageENS.vue
```diff
@@ -566,37 +514,6 @@ export default {
       }
     },
     processPermanentRegistrar() {},
```
kvhnuke: empty `function`

src/dapps/ManageENS/ABI/registrarAbi.js
gesseekur: @kvhnuke It does look like its been referenced on ManageENS.vue.341

src/dapps/ManageENS/ManageENS.vue
gesseekur: removed!

src/dapps/BuySubdomain/BuySubdomain.vue
gesseekur: changed 👍

PR #1433, opened 2019-09-16
bitbox-windows: fix u2f timeout for windows users

Tomasvrba:
Increases timeout to 35 seconds for Windows users instead of polling
every 3 seconds since Windows 10 1903+ waits for user confirmation

### Bug
- Fixes U2F hijack issue for new Windows 10 update 1903 by waiting 35 seconds (to accommodate 30 second firmware timeout) instead of polling every 3 seconds. Windows < 10, MacOS and Linux are not affected, this is a Windows 10 only fix. 
- Windows 10 users pre-1903 update may have to wait up to 25 seconds if they do not sign a transaction within 3 seconds. 
- Fix tested on newest Chrome on Windows 10 1903, Windows 10 1803, Ubuntu 18.04 and MacOS

@SteveMieskoski 
- There is an associated firmware patch release v7.0.1 here: https://github.com/digitalbitbox/mcu/releases/tag/v7.0.1 that can be loaded via the `load_firmware.py` script for testing
- **The firmware will be released to users tomorrow, Sep 17th for updating via the BitBox App**
- The applied fix can be tested on our instance of MEW at: https://digitalbitbox.com/mew/

### DIFF
--- a/src/wallets/hardware/bitbox/digitalBitboxEth.js
+++ b/src/wallets/hardware/bitbox/digitalBitboxEth.js
@@ -12,7 +12,9 @@
 
 import * as Crypto from 'crypto';
 import * as HDKey from 'hdkey';
-const BitBoxSupportedMajorVersion = 7;
+import * as semver from 'semver';
+
+const BitBoxSupportedVersion = '7.0.1';
 const hijackState = {
   // Order must match that in the firmware code
   responseReady: 0,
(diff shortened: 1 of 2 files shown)


SteveMieskoski:
@Tomasvrba I updated the firmware and we tried the mew version from you guys on two windows machines here and received this windows popup message. 
photo_2019-09-16_17-45-55


Do you have an idea what could be the cause?   I figured I would ask before digging into this.
Thanks!

Tomasvrba:
Hi @SteveMieskoski, thanks for checking about the popup. That's now standard Windows behavior since the new 1903 update. Windows now has its own FIDO2/U2F implementation and all U2F communication on Windows now goes through the OS instead of the u2f-api in the browser, so Windows essentially takes over. 
For every u2f-sign/authenticate/register request Windows now shows this popup.

SteveMieskoski:
Hi @Tomasvrba, thanks for getting back so quick.  And, yes, on the one machine we were not able to get past it to sign into the wallet interface.  The other two showed the popup, but after a moment the device returned it's information and then the site progressed to the address selection modal.

Tomasvrba:
And that happens reproducibly every time?
Could you share what version and build of Windows those machines were running to see if we can reproduce?

PR #1442, opened 2019-09-19
Dev/mew decision tree new

hsurf22:
### Feature

- [x] Updated CHANGELOG.md
- [x] Add PR label
  - Add decision tree

https://www.youtube.com/watch?v=U4tEnBwLbc0&feature=youtu.be

Screenshot from 2019-09-18 21-26-52


SteveMieskoski:
Also, noticed the 'house' home icon on the bottom has hover effects, but doesn't have a click event. (does not do anything when clicked)

hsurf22:
> This should probably show in all pages, not just the home page.

Yes. I think this is showing on all other pages.
I just tested it and works on all pages.

hsurf22:
> Also, noticed the 'house' home icon on the bottom has hover effects, but doesn't have a click event. (does not do anything when clicked)

This button brings user to the very top of the decision tree category.
Now home button is hidden when user is on top category, so users can't see the button.

hsurf22:
Markup data has been updated
and some design changes have been made.

PR #1519, opened 2019-10-16
App optimization 

gesseekur:
### Devop

- [x] Updated CHANGELOG.md
- [x] Add PR label


src/components/Inputs/InputSearch/InputSearch.vue
```diff
@@ -1,27 +1,12 @@
...
   <div class="input-search">
     <slot />
     <img src="@/assets/images/icons/magnifier.svg" alt="Search" />
   </div>
 </template>
 
 <script>
-export default {
-  props: {
-    options: {
```
gesseekur: doesn't look like this props is being used anywhere

src/components/DecisionTree/components/SearchContainer/SearchContainer.vue
```diff
@@ -1,4 +1,4 @@
-<template>
+<template functional>
```
gesseekur: I couldn't find where this is being used...not sure if we should remove it

gesseekur: removed for now

src/components/Inputs/InputSearch/InputSearch.vue
gesseekur: removed for now

PR #1521, opened 2019-10-18
Feature/ens multi coin support

kvhnuke:
https://eips.ethereum.org/EIPS/eip-2304

### Feature

- [x] Updated CHANGELOG.md
- [x] Add PR label


src/layouts/InterfaceLayout/containers/SwapContainer/components/SwapAddressSelector/SwapAddressSelector.vue
```diff
@@ -71,12 +71,12 @@
               <p class="listed-address">
                 {{ addr.address }}
                 <!-- Address book feature
-                <span
-                  v-if="addr.address !== currentAddress && addr.currency !== 'ETH'"
-                  class="address-note"
-                  >{{ addr.currency }} {{ $t('interface.addr') }}</span
-                >
-                -->
+                                <span
```
kvhnuke: @SteveMieskoski are these comments necessary?

SteveMieskoski: I don't think so.  It was back when it looked like we were talking about adding an address book feature in the near future so I kept them in there, but that is a while back now.   I'm going to remove them.

kvhnuke: please do

PR #1548, opened 2019-10-22
Update egem.js

r2clan:
Please update network for egem old url no longer works.


gamalielhere:
Matches listed in official wiki https://wiki.egem.io/en/egemwallet#connect-2

PR #1573, opened 2019-10-28
Fix/ENS empty address

kvhnuke:
### Bug

- [x] Updated CHANGELOG.md
  - [x] Link to issue: https://discuss.ens.domains/t/new-standard-proposal-ens-multicoin-support/1148/46
- [x] Add PR label

### DIFF
--- a/src/dapps/ManageENS/ManageENS.vue
+++ b/src/dapps/ManageENS/ManageENS.vue
@@ -252,12 +252,19 @@ export default {
         ResolverAbi,
         publicResolverAddress
       );
+      const getDecodedAddress = _coinItem => {
+        let decodedAddress = '0x';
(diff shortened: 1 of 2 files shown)


src/dapps/ManageENS/ManageENS.vue
```diff
@@ -252,12 +252,19 @@ export default {
         ResolverAbi,
         publicResolverAddress
       );
+      const getDecodedAddress = _coinItem => {
+        let decodedAddress = '0x';
```
makoto: Should be either `Buffer.from("")` or `new Buffer(0)` to set empty string. cc @Arachnid

kvhnuke: Abi encoder converts '0x' to an empty buffer.

PR #1741, opened 2019-12-11
Bug/chainid for send offline helper

SteveMieskoski:
### Bug

- [x] Updated CHANGELOG.md
- [x] Add PR label

- Fix incorrect chain-id breaking offline send helper
- Fix out of memory error on build


package.json
```diff
@@ -11,7 +11,7 @@
     "build:mewcx": "ROUTER_MODE=hash BUILD_TYPE=mewcx npm run build",
     "start:mewcx": "ROUTER_MODE=hash BUILD_TYPE=mewcx npm run buildwatch",
     "buildwatch": "npm run update && vue-cli-service build --watch",
-    "build": "npm run update && vue-cli-service build",
+    "build": "npm run update && NODE_OPTIONS='--max-old-space-size=4092' vue-cli-service build",
```
kvhnuke: we cannot do this cz the travis build sets up the proper node_options
you can let the user know they can simply run `NODE_OPTIONS=--max-old-space-size=4092 npm run build:hash`

PR #1755, opened 2019-12-16
Devop/notification pruning

SteveMieskoski:
### Devop

- [x] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings

- Prune notifications keeping those less than five days old or the 25 most recent


src/containers/NotificationsContainer/NotificationsContainer.vue
```diff
@@ -171,6 +171,12 @@ export default {
   watch: {
     notifications() {
       this.countUnread();
+    },
+    network() {
+      this.pruneNotifications();
+    },
+    ['account.address']() {
+      this.pruneNotifications();
```
kvhnuke: Prune whenever there is a new notification, not when it changes the network or address

PR #1795, opened 2019-12-23
Feature/update address book

gesseekur:
### Feature

- [x] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings


src/components/AddressBookModal/AddressBookModal.vue
```diff
@@ -0,0 +1,217 @@
...
+    return {
+      contactNickname: '',
+      contactAddress: '',
+      isValidAddress: false
+    };
+  },
+  computed: {
+    ...mapState(['addressBook']),
+    isBtnDisabled() {
+      if (this.title === 'Edit Address' && this.addressBook[this.currentIdx]) {
```
kvhnuke: Don't check against strings, export an enum
```
const seasons = {
    SUMMER: 'summer',
    WINTER: 'winter',
    SPRING: 'spring',
    AUTUMN: 'autumn'
}

season === seasons.SPRING
```
Then you can use it on both here and wherever you pass the prop

gesseekur: 👍 👍

PR #1809, opened 2020-01-02
Feature/unstoppableResolution

JohnnyJumper:
### Devop

- [ ] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings


src/directives/AddrResolver.js
```diff
@@ -1,12 +1,8 @@
-import normalise from '@/helpers/normalise';
-import { Misc } from '@/helpers';
```
gamalielhere: This is needed still

src/directives/AddrResolver.js
```diff
@@ -1,12 +1,8 @@
...
 import { toChecksumAddress } from '@/helpers/addressUtils';
-import utils from 'web3-utils';
-import WAValidator from 'wallet-address-validator';
```
gamalielhere: Same goes for this

src/directives/AddrResolver.js
gamalielhere: This is still being used in the code.

src/directives/AddrResolver.js
```diff
@@ -34,157 +30,65 @@ const AddrResolver = {
...
-            const isValid = MAValidator.validate(e, parentCurrency);
-            if (isValid) {
-              _this.isValidAddress = isValid;
-              _this.hexAddress = e;
-            } else {
-              _this.isValidAddress = false;
-              _this.hexAddress = '';
+      try {
+        const address = await resolution.addressOrThrow(domain, parentCurrency);
+        if (address) {
```
gamalielhere: Might also want to add an error similar to what you did on [here](https://github.com/MyEtherWallet/MyEtherWallet/pull/1809/files#diff-25ab0cdc1d5941225b37bcab2160e95dR89) in the instance that it returns a blank address which I'm assuming it shouldn't

src/directives/AddrResolver.js
```diff
@@ -34,157 +30,65 @@ const AddrResolver = {
...
-                  );
-                }
-              } else {
-                errorPar.innerText = '';
-              }
-
-              el.parentNode.parentNode.appendChild(errorPar);
+              messagePar.classList.add('resolver-addr');
+              _this.isValidAddress = true;
+              _this.hexAddress = toChecksumAddress(address);
```
gamalielhere: `_this.isValidAddress = true;` and `_this.hexAddress = toChecksumAddress(address);` conflicts with the same assignment on line 66 and 67. These should be an else on this if statement or the other way around.

gamalielhere:
Aside from this, everything looks good! @JohnnyJumper 

Screen Shot 2020-01-02 at 2 04 04 PM

JohnnyJumper:
@gamalielhere  should be good to go now. I did a small re-factor and made sure that the error message will disappear once the valid address is set, or the text field is empty.

gesseekur:
Hm! I am getting this error on the swap page for this branch , it doesn't seem to be happening on master. 
Screen Shot 2020-01-03 at 11 42 26 AM
Screen Shot 2020-01-03 at 11 42 38 AM

SteveMieskoski:
@JohnnyJumper I was looking over the source of your npm package and noticed it may not include all the networks where ENS names are present. Is this correct and do you plan to add them (for example gorlil).  Also, please use the network and url passed from the network field in the Vuex store to identify the node/endpoint which should be used when interacting with the blockchain.  Thanks!

JohnnyJumper:
## @gesseekur 
Looks like something in WAValidator won't pass the address. I removed the check and it seems to be working on both pages SWAP and SEND, tested with different domains: jessica.chainlinknode.eth brad.zil brad.crypto mewtopiax.eth looks like no more issues.

Please take a look maybe you will be able to spot some more issues? 

## @SteveMieskoski
I used the network from the vuexStore. If you can give me the addresses of different test networks for ENS I can add them very easily to our package.

Thank you very much for testing this and helping me out with the issues, very much appreciated it!

gamalielhere:
Moved to #1857

Issue #1847, opened 2020-01-12
Users do not know to expect the build to fail when new dependency versions are released, and the errors don't look like errors or suggest a solution.


interfect:
- **I'm submitting a ...**

  - [ ] Feature request
  - [X] Bug report

- **Bug Report**
  - What is the current behavior?
    - As MEW's dependencies release new versions, and those release dates move further into the past, previously-working commits of MEW stop being able to build and become broken. This is enforced by the `package-test.js` script, which gets run as one of the first parts of `npm run build`. However, when the build fails for this reason, it is unclear exactly what the problem is, or how to fix it. The [Getting Started instructions](https://github.com/MyEtherWallet/MyEtherWallet#getting-started) do not mention this kind of failure as a possible outcome of `npm run build`, even though the current master branch (e451ff50a30bf29a307529ede391caf9dc965478) exhibits this problem as of this writing. Moreover, the error messages that announce the dependency updates don't really look like errors; they just say `new update found` and some information about the versions. It is not really clear that the existence of the new version is an error, nor that the build process expects manual intervention from the user in `package.json` in order to fix it.
  - If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem
    - Have a working MEW build environment
    - Check out current master (e451ff50a30bf29a307529ede391caf9dc965478)
    - `rm -Rf node_modules; npm install`
    - `npm run build`
  - What is the expected behavior?
    - IMHO, previously-working builds should not start to fail either because of a new release of a dependency or the passage of time, especially if there is not an automated process in place to keep the current master branch building. But if it is important for security reasons to stop users from downloading and building MEW on top of dependencies for which newer versions have been published, then when that condition occurs the script needs to instruct the users about why they are not permitted to continue with their build, and what they are expected to do about it (i.e. go hack `package.json`, or wait for an update to MEW).
  - What is the motivation / use case for changing the behavior?
    - I downloaded MEW to build it myself, so I could take advantage of its ability to run off my own machine, and following the Getting Started instructions didn't work. I then had to dig through the build system and use my esoteric knowledge as a vaguely competent nodejs user to work out that the build system was intentionally stopping. Users shouldn't have to do that.
  - Please tell us about your environment:
    - Operating system
      - Ubuntu 18.04
    - Browser and version
      - N/A
  - Other information (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)
[...]

PR #1848, opened 2020-01-12
Errorize out of date package errors

interfect:
This is in reference to #1847. It makes the errors that show up in the build system when dependencies release new versions look like actual errors, and adds a note about how you are supposed to fix them.

### Bug

- [x] Updated CHANGELOG.md
- [X] Is this a user submitted bug?
  - [X] Link to issue: https://github.com/MyEtherWallet/MyEtherWallet/issues/1847
- [x] Add PR label **I cannot do this as I lack permissions**
- [X] No hard-coded strings
  - At least not in the app itself. The build system is full of them.


gamalielhere:
@interfect you should be able to pull down develop and fix the build now

PR #2046, opened 2020-03-02
Fix Buy Eth Widget bug [v5]

SteveMieskoski:
### Bug

- [x] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings

- fix error coming from not getting the max/min using currency
- Fixes Buy Eth Widget bug posted in asana


gesseekur:
I think you have to add 'decompress' to AUDIT_EXCEPTIONS in package-audit.js to fix the build!

PR #2070, opened 2020-03-09
Add Unstoppable Dapp

sudoryan:
Add Unstoppable Domains dapp for selling .crypto domains via Stripe.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -50,6 +50,7 @@
       ]
       }
     </script>
+  <script src="https://js.stripe.com/v3/"></script>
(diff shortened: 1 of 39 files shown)


public/index.html
```diff
@@ -50,6 +50,7 @@
       ]
       }
     </script>
+  <script src="https://js.stripe.com/v3/"></script>
```
kvhnuke: can this be a npm package? or else we have to save it to a local file, we done allow loading scripts from 3rd party domains

sudoryan: Sure. I've added the @stripe/stripe-js dependency that does this.

kvhnuke: Thanks

PR #2158, opened 2020-03-25
Tomasvrba feature/bitbox02 integration

kvhnuke:
### Devop

- [x] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings


Tomasvrba:
@kvhnuke or @SteveMieskoski 
Please merge https://github.com/MyEtherWallet/MyEtherWallet/pull/2163 before this PR

Tomasvrba:
@kvhnuke Pasting this here from the other PR which had been merge so this doesn't get lost, with regards to the library size:

> Thanks @Tomasvrba for the fix, we will include this. PR significantly increased final build size and broke the travis build due to virustotal limitations. so we had to rebase the code to seperate it, that is probably when the we lost some of the code. Since then we made the build pass, but looking for other ways to reduce the final build size. We will include the PR in the release, however let us know if you know anyway to split the chunks more efficiently to reduce overall build size

Great, thank you! Yeah, I mentioned the size to @SteveMieskoski in an email a while ago and also on the original PR. One way to address it on MEW side would be to split the large bitbox02-api dependency into a separate chunk with webpack instead of bundling all node_modules into a single `vendors` chunk as configure in `defaultConfigs.js`

```
  optimization: {
    splitChunks: {
      cacheGroups: {
        commons: {
          test: /[\\/]node_modules[\\/]/,
          name: 'vendors',
          chunks: 'initial'
        }
      }
    }
}
```

This would then let us lazy-load the library only on demand when the user actually wants to use the BB02. 
I played around with this a little bit when doing the integration, but as mentioned on the original PR, I didn't want to mess with your configs before discussing this. 
Here are some references that I was using for dynamic webpack loading/vue lazy-loading:

https://webpack.js.org/guides/lazy-loading/
https://vuedose.tips/tips/dynamic-imports-in-vue-js-for-better-performance/
https://vueschool.io/articles/vuejs-tutorials/lazy-loading-individual-vue-components-and-prefetching/

Let me know if that's something you'd consider. 

On our side, we are also of course working on trying to optimize GopherJS to shrink the final size of the js library compiled from Go.

Issue #2052, opened 2020-03-03
Access hardwareWallet object instance in InterfaceLayout.vue (for switchAddress())


Tomasvrba:
- **I'm submitting a ...**

  - [x] Feature request

- **Feature Request**

  - Briefly explain the feature you would like us to implement

Instead of re-creating the hww object in `switchAddress()`, be able to access the existing object in `InterfaceLayout.vue` with `this.hardwareWallet` like already possible in `NetworkAndAddressModal.vue` on the props:
```
  props: {
    hardwareWallet: {
      type: Object,
      default: function() {
        return {};
      }
    }
```

  - What are the current issues this feature will solve

The BitBoxBridge is able to detect when a connection closes, so I would like to provide it with the `clearWallet` action as a callback, to clear the MEW interface when the user unplugs the usb device. However, this causes an issue when switching addresses, as the entire hww object is re-created, a new connection is initialized, the old websocket closes and therefore `clearWallet` is called. When this happens, then `toggleNetworkAddrModal(walletInstance)` no longer has access to the necessary `$refs` after `clearWallet`. 

It would be great to just be able to call `this.hardwareWallet.init(selectedPath)` on the existing hww instance like in `NetworkAndAddressModal.vue`, then on the hww side, I can implement `init` to not close the websocket connection to the BitBoxBridge.

  - Is this feature cross compatible with other ethereum networks

Yes
  - Are you willing to help us with development support if required

Yes
  - Where can this be added?

`/interface`

kvhnuke:
I believe this is a good idea @gamalielhere @SteveMieskoski thoughts?

PR #2163, opened 2020-03-26
Bitbox02 integration fix

Tomasvrba:
### Feature

Update bitbox api js to 0.4.0

### Bug

Re-add lost bitbox modals

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32414,9 +32414,9 @@
       }
     },
     "bitbox02-api": {
-      "version": "0.3.0",
-      "resolved": "https://registry.npmjs.org/bitbox02-api/-/bitbox02-api-0.3.0.tgz",
-      "integrity": "sha512-vvGB6vvR86KQcXJH7UmYrbFXdfnDevEomVCZanGxPcT7fj5M15fSF47smyAi8ySYmM9yX4DtXq/cmjtLNK+hbQ==",
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/bitbox02-api/-/bitbox02-api-0.4.0.tgz",
+      "integrity": "sha512-fMGc8Hfx6HlCL6fCx4VBo4NyMYvyrszaTxuwFQQys9Ewem/SWuvk8ZgLuuw9yZKdXIxEvlEAHc2e5wStI83p9g==",
       "dev": true
     },
     "bl": {
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "bc-vault-js": "2.0.3",
     "bignumber.js": "9.0.0",
     "bip39": "3.0.2",
-    "bitbox02-api": "0.3.0",
+    "bitbox02-api": "0.4.0",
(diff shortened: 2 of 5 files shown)


Tomasvrba:
@kvhnuke Please see commit message: https://github.com/MyEtherWallet/MyEtherWallet/pull/2163/commits/95ee18a768d2534aafb1480dcf00f4454577ad2b

Seems like the BitBox modals from my previous PR had been lost during refactoring of the AccessWalletLayout, so I'm re-adding them here.

Also updating the lib to the newly released 0.4.0 and firmware 7.0.0.
@SteveMieskoski To test, you can update to the newest firmware with the current 4.18.0 app: https://shiftcrypto.ch/start/

Tomasvrba:
By the way, thank you @kvhnuke for fixing the store dispatch code in the previous PR. Clearing the wallet on device unplug now works great. 

Could I please get a comment from you on this related issue? https://github.com/MyEtherWallet/MyEtherWallet/issues/2052

Because currently with using clearWallet, I don't know how to make address switching work. Details are in the issue.

kvhnuke:
Thanks @Tomasvrba for the fix, we will include this. PR significantly increased final build size and broke the travis build due to virustotal limitations. so we had to rebase the code to seperate it, that is probably when the we lost some of the code. Since then we made the build pass, but looking for other ways to reduce the final build size. We will include the PR in the release, however let us know if you know anyway to split the chunks more efficiently to reduce overall build size

Tomasvrba:
> Thanks @Tomasvrba for the fix, we will include this. PR significantly increased final build size and broke the travis build due to virustotal limitations. so we had to rebase the code to seperate it, that is probably when the we lost some of the code. Since then we made the build pass, but looking for other ways to reduce the final build size. We will include the PR in the release, however let us know if you know anyway to split the chunks more efficiently to reduce overall build size

Great, thank you! Yeah, I mentioned the size to @SteveMieskoski in an email a while ago and also on the original PR. One way to address it on MEW side would be to split the large bitbox02-api dependency into a separate chunk with webpack instead of bundling all node_modules into a single `vendors` chunk as configure in `defaultConfigs.js`

```
  optimization: {
    splitChunks: {
      cacheGroups: {
        commons: {
          test: /[\\/]node_modules[\\/]/,
          name: 'vendors',
          chunks: 'initial'
        }
      }
    }
}
```

This would then let us lazy-load the library only on demand when the user actually wants to use the BB02. 
I played around with this a little bit when doing the integration, but as mentioned on the original PR, I didn't want to mess with your configs before discussing this. 
Here are some references that I was using for dynamic webpack loading/vue lazy-loading:

https://webpack.js.org/guides/lazy-loading/
https://vuedose.tips/tips/dynamic-imports-in-vue-js-for-better-performance/
https://vueschool.io/articles/vuejs-tutorials/lazy-loading-individual-vue-components-and-prefetching/

Let me know if that's something you'd consider. 

On our side, we are also of course working on trying to optimize GopherJS to shrink the final size of the js library compiled from Go.

PR #2309, opened 2020-04-27
Add Unstoppable Domain Management

sudoryan:


kvhnuke:
Hi @sudoryan can you fix the build, I think you just have to remove the console.log
also please remove the yarn.lock we use package-lock

sudoryan:
> Hi @sudoryan can you fix the build, I think you just have to remove the console.log
> also please remove the yarn.lock we use package-lock

Thanks for taking a look! Yarn.lock has been removed as well as the console.log

PR #2323, opened 2020-04-28
polishing refinements to dex.ag

SteveMieskoski:
### Devop

- [ ] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings


src/partners/dexAg/dexAg.js
```diff
@@ -313,6 +314,17 @@ export default class DexAg {
         data: tradeDetails.trade.data,
         value: tradeDetails.trade.value
       };
+      if (
+        tradeDetails.metadata.gasPrice &&
+        swapDetails.provider === 'bancor' &&
```
gesseekur, line 319: I think we don't like to compare values to strings ??

PR #2338, opened 2020-05-01
Feature/combine tb contract with local

gamalielhere:
### Feature

- [x] Updated CHANGELOG.md
- [x] Add test cases or procedure
- [x] Add PR label
- [x] No hard-coded strings


src/layouts/InterfaceLayout/InterfaceLayout.vue
```diff
@@ -437,34 +437,41 @@ export default {
...
+            });
+
+            if (!found) return token;
+          });
+
+          tokens = tokens.concat(filteredNetwork);
         } catch (e) {
           tokens = this.network.type.tokens.map(token => {
-            token.balance = 'Load';
+            token.balance = 0;
```
kvhnuke, line 468: Shouldn't this still be Load? On network error they should be able to manually Load

gamalielhere: right. I'll change it

PR #2380, opened 2020-05-16
supply buffers to generate address when deploying a contract

SteveMieskoski:
### Bug

- [x] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings

-  ethereum-utils generateAddress function takes buffers as arguments.  Previously strings were supplied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 - Remove CW promo [#2377](https://github.com/MyEtherWallet/MyEtherWallet/pull/2377)
 
+### Bug
+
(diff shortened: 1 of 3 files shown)


kvhnuke:
I added this here, https://github.com/MyEtherWallet/MyEtherWallet/pull/2338/files#diff-ba2985346781a84c193a54d3dfb284c2R326

also `Buffer.from()` will simply convert utf string to buffer it wont take '0x' into consideration and assume it is a hex string, `toBuffer` from `web3-utils` takes '0x' into consideration

PR #2389, opened 2020-05-20
bug/perkle [v5]

gamalielhere:
### Bug

- [x] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings


alanwilhelm:
Did this make it into the release? We are seeing connection problems for PRKL on the live site. The XSS wants to allow port 443 but the service is still trying to connect to 8501. Perhaps this change hasn't made it to the release as it looks like the latest update changed both settings to 8501 and we are still seeing https://perkle-pub.esprezzo.io:443 in the javascript XSS allowed domains on the live site. Both 443 and 8501 work using https on our public node. It looks like they just need to match in the release code as right now `connections.js` is allowing 443 and `src/networks/nodes/prkl.js` is trying to connect to 8501. Let me know if I should submit a pull request.

gamalielhere:
@alanwilhelm it should be working now

PR #2397, opened 2020-05-22
Bug/unable to validate address

SteveMieskoski:
### Bug

- [x] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings


gesseekur:
wasn't tooo sure how to double check this on the build but I was able to do a swap successfully

src/directives/AddrResolver.js
```diff
@@ -173,6 +173,7 @@ const AddrResolver = {
         } catch (e) {
           if (e.message.includes('Missing validator for currency: ')) {
             _this.isValidAddress = true;
+            _this.hexAddress = domain;
```
kvhnuke, line 176: `_this.hexAddress` should always be the hexAddress of the resolved name

SteveMieskoski: it wasn't getting set when it wasn't a domain name, and was unable to validate.  In which case the domain is the address passed into the resolver/ validator

PR #2444, opened 2020-06-15
Unstoppabledomains/support for dash domains

JohnnyJumper:
* Updated the regex pattern to support domains with - and numbers
* turned off ens from lib since it is not being used


gamalielhere:
moved to #2456

PR #2484, opened 2020-06-24
add check for market impact when finalizing swap details

SteveMieskoski:
### Devop

- [x] Updated CHANGELOG.md
- [x] Add PR label
- [x] No hard-coded strings


src/partners/dexAg/dexAg.js
```diff
@@ -376,6 +376,11 @@ export default class DexAg {
       : 'ag';
 
     const tradeDetails = await this.createTransaction(swapDetails, dexToUse);
+    if (tradeDetails.metadata.marketImpact) {
```
kvhnuke, line 379: `tradeDetails.metadata.marketImpact` this always exists
we have to make sure market impact is less than 10

PR #2630, opened 2020-10-18
package.json: update bitbox02 to 0.7.0

benma:
Other deps were also bumped according to the error message of `npm run
build`.


gamalielhere:
Included in #2634

PR #2647, opened 2020-10-27
Improve start script (simplicity, performance, user-friendliness, reproducibility)

theqp:
Problems with the current start script I have solved with this one-liner:

- `npm run start` does the same thing as `dev` (just slower because runs `update:mainlists` twice)
- is not reproducible 
  - `package-test.js` might fail even on release commits
  - changes files which are in the repository (`fetchLists/lists`)
- slower than it should be, scripts which make the build not reproducible take a long time
  ```
  node package-test.js
  ...
  node package-test.js  4.08s user 0.45s system 1% cpu 6:11.40 total
  ```
  6 minutes :watch: 
- makes starting the program cumbersome (why edit `package.json` manually just to try out a version?)

Sorry for sounding condescending, there may be a reasoning behind your solutions, I accept the same treatment :grinning:


theqp:
Build fails because of `package-test.js`. Ironic.

kvhnuke:
Im bit confused what are you trying to accomplish?

kvhnuke:
you are right about list updating twice `"start": "npm run dev"` should fix it, can you make that change to your PR

theqp:
> Im bit confused what are you trying to accomplish?

I want to start the program without getting any errors (caused by `package-test.js`) or any files modified which are in the repo (e.g. `fetchLists/lists`).

kvhnuke:
> > Im bit confused what are you trying to accomplish?
> 
> @kvhnuke I want to start the program without getting any errors (caused by `package-test.js`) or any files modified which are in the repo (e.g. `fetchLists/lists`).
> 
> I have updated `start` as you suggested.

got it! your `start:ci` definitely accomplish this

kvhnuke (approved):

PR #2764, opened 2021-01-04
Forkwatch ninja.besu patch

forkwatch-ninja:
## Fix

Remove defunct Websocket URL for ETC and rename other ETC Coop node to Besu (because it's using Hyperledger Besu Node)


bobsummerwill:
Hey guys!
This PR in response to reported issues with ETC on MEW.
Looks like we had a "dangling" option for WebSockets which will not work with ethercluster.
CC @stevanlohja.

kvhnuke (approved):

bobsummerwill:
Thanks everyone!

PR #2848, opened 2021-02-19
V6/fix wallet layouts

hsurf22:
### Devop

* \[x] Updated CHANGELOG.md
* \[x] Add PR label


src/views/layouts-wallet/TheSendTransactionLayout.vue
```diff
@@ -1,10 +1,9 @@
 <template>
   <v-row>
-    <v-col cols="9">
+    <v-col cols="8" lg="8">
```
olgakup: so cols = 8 meaning that on all views  it will take 8 cols of 12, you dont have to specify lg = 8, unless you want to change how many column its suppose to take up starting from that break point.

src/views/layouts-wallet/TheSendTransactionLayout.vue
```diff
@@ -1,10 +1,9 @@
...
       <module-send />
     </v-col>
-    <v-spacer cols="1" />
-    <v-col cols="3">
+    <v-col cols="4" lg="4">
```
olgakup: the same thing here with cols

src/views/layouts-wallet/TheDashboardLayout.vue
```diff
@@ -158,7 +158,7 @@
         </mew6-white-sheet>
       </div>
     </v-col>
-    <v-col cols="4" lg="4">
+    <v-col cols="12" md="4">
```
olgakup, line 161: I think the design has this on top, can you add the oder to it?

src/views/layouts-wallet/TheSendTransactionLayout.vue
hsurf22: Fixed
https://github.com/MyEtherWallet/MyEtherWallet/blob/v6/fix-wallet-layouts/src/views/layouts-wallet/TheSendTransactionLayout.vue

gamalielhere (approved):

PR #2886, opened 2021-03-09
fix: langing page header

hsurf22:
### Devop

* \[x] Updated CHANGELOG.md
* \[x] Add PR label
   - Fix landing page header layout
   - Fix mew tools menu


gamalielhere (approved):

src/views/components-default/TheDefaultHeader.vue
```diff
@@ -36,34 +36,16 @@
...
+          <mew-tools class="ml-auto" />
         </v-col>
       </v-row>
     </v-container>
   </div>
 </template>
 
 <script>
-import mewTools from '@/components/mewTools/MewTools';
+import mewTools from '@/components/mew-tools/MewTools';
```
olgakup, line 48: When you import ui components please keep the same name casing so its 
MewTools

olgakup (changes requested):

src/views/components-default/TheDefaultHeader.vue
hsurf22: Got it!!

PR #2915, opened 2021-03-19
updating module token

gesseekur:
### Devop

* \[x] Add PR label


src/modules/balance/ModuleTokens.vue
```diff
@@ -75,46 +103,47 @@ export default {
         })
         .map(item => {
           const newObj = {};
-          newObj.value = `$ ${item.usdBalance}`;
+          newObj.balance = [
+            new BigNumber(
+              this.web3.utils.fromWei(item.balance, 'ether')
```
gamalielhere, line 108: can you import fromWei from utils instead of importing web3? otherwise looks good

gamalielhere (approved):

src/modules/balance/ModuleTokens.vue
gesseekur: done!

olgakup (approved):

PR #2945, opened 2021-03-30
unstoppable dapp

JohnnyJumper:
This PR is for integration Unstoppable Domains dapp into dapp center


JohnnyJumper:
The application is ready, however, I found a very strange issue with using web3.eth.sendTransaction that I might need someone's help. 

Whenever I try to send a transaction I am getting a `Given input "NaN" is not a number.`

I have double-checked all the data I am sending and it seems to be correct. Here is the stack trace I am getting 
```
Error: Given input "NaN" is not a number.
    at Object.numberToHex (utils.js:238)
    at formatters.js:144
    at Array.forEach (<anonymous>)
    at _txInputFormatter (formatters.js:143)
    at Method.inputCallFormatter (formatters.js:156)
    at index.js:130
    at Array.map (<anonymous>)
    at Method.push../node_modules/web3-core-method/lib/index.js.Method.formatInput (index.js:128)
    at Method.push../node_modules/web3-core-method/lib/index.js.Method.toPayload (index.js:161)
    at Object.send [as estimateGas] (index.js:499)
    at _callee$ (app.js:82630)
    at tryCatch (runtime.js:64)
    at Generator.invoke [as _invoke] (runtime.js:294)
    at Generator.next (runtime.js:119)
    at asyncGeneratorStep (_index.mjs:3)
    at _next (_index.mjs:25)
```

You can replicate this by trying to buy a domain with crypto even if you have not enough balance. I have left the button clickable to make it simple to catch the bug. It is not related to not enough funds, as I have checked it with several wallets with enough funds on each. 

If someone can take a quick look to spot what can be an issue.

The fault code can be found at UnstoppableDomainBuyOverlay.vue at line 362.


For reference, this is the [old code](https://github.com/MyEtherWallet/MyEtherWallet/blob/2036a5f7b22067d1bb083c95e88c0976939b250d/src/dapps/Unstoppable/containers/CryptoContainer/CryptoContainer.vue#L279-L293) from old integration that has been working so far.

src/dapps/unstoppable-domains/components/UnstoppableDomainBuyOverlay.vue
```diff
@@ -136,9 +199,191 @@ export default {
...
+      'getOrder'
+    ]),
+    ...mapMutations('unstoppable', ['setOrder']),
+    ...mapActions('unstoppable', ['updateOrderStatus']),
+    selectTab(event) {
+      const tabId = event.currentTarget.id;
+      this.crypto = tabId === 'crypto';
+      this.credit = !this.crypto;
+    },
+    async getEthPrice() {
```
gamalielhere, line 299: You can replace this by importing eth value from external store:
`...mapState('external', ['ETHUSDValue']),` and replacing line 223-226 with:
```
ethConvertPrice = new BigNumber(domainPrice)
          .dividedBy(this.ETHUSDValue.value)
          .toFixed(8);
      }
```

PR #2998, opened 2021-04-06
V6/fix mobile menu logo and max screen size

hsurf22:
### Bug

* \[x] Updated CHANGELOG.md
* \[x] Add PR label
   - Fix mobile menu logo
   - Fix max screen width for desktop

Screenshot from 2021-04-06 01-03-42

Screenshot from 2021-04-06 01-04-02


gamalielhere (approved):

src/assets/styles/GlobalStyles.scss
```diff
@@ -6,6 +6,12 @@
   }
 }
 
+@media (min-width: 1904px) {
+  .container {
```
olgakup, line 10: What is this guy for?

olgakup: if you are adding the max width for the whole view, please use it simularly as @media 
 wallet-content-container, and not container

olgakup (changes requested):

src/assets/styles/GlobalStyles.scss
hsurf22: I understand what you mean, but I'm sure this is the right way to limit the web app width, because we use v-container in almost every components to limit their width.
There's no reason to create separate container class to limit the web app width.
<container> is made for this purpose.

PR #3104, opened 2021-05-25
V6/swap button switch

gamalielhere:
### Devop

* \[x] Updated CHANGELOG.md
* \[x] Add PR label
* \[x] No hard-coded strings


src/modules/swap/ModuleSwap.vue
```diff
@@ -47,7 +47,13 @@
             /></v-col>
 
             <v-col cols="12" sm="2" class="pt-0 pt-sm-3">
-              <div class="d-flex align-center justify-center">
+              <div
+                :class="[
```
olgakup, line 51: add disabled computed property on it:

disabled if reverse pair does not exists

olgakup (changes requested):

olgakup (approved):

PR #3107, opened 2021-05-26
V6/swap remove to from vice versa

gamalielhere:
### Devop

* \[x] Updated CHANGELOG.md
* \[x] Add PR label
* \[x] No hard-coded strings


src/modules/swap/ModuleSwap.vue
```diff
@@ -697,6 +739,12 @@ export default {
     buyEth() {
       window.open('https://ccswap.myetherwallet.com/#/', '_blank');
     },
+    switchTokens() {
```
olgakup, line 742: right now wen you click switch To field becomes empty. Is should have become the original From token

olgakup (changes requested):

src/modules/swap/ModuleSwap.vue
olgakup: oh never mind there is a different pr for that

olgakup (approved):

PR #3112, opened 2021-05-27
V6/number formatter

olgakup:
* \[x] Add PR label


gamalielhere (approved):

src/dapps/ens-manager-dapp/components/manage/ManageRenew.vue
```diff
@@ -22,6 +22,10 @@
 </template>
 
 <script>
+import {
+  formatFloatingPointValue,
+  formatFiatValue
+} from '../../../../core/helpers/numberFormatHelper';
```
gesseekur, line 28: will it be better to just do @/core/ ?

gesseekur (approved):

src/dapps/ens-manager-dapp/components/manage/ManageRenew.vue
olgakup: yes it will be. I will fix that

PR #3133, opened 2021-06-08
devop: 🔧 add aave apollo mixin

gesseekur:
Aave apollo graphql refactor


src/dapps/aave-dapp/components/AaveAmountForm.vue
```diff
@@ -142,11 +139,18 @@ export default {
   computed: {
     hasAmount() {
       return BigNumber(this.amount).gt(0);
+    },
+    checkIfNumerical() {
+      const regex = new RegExp('^-?[0-9]+.?[0-9]*$');
```
rsteubs, line 144: This regex would accept values like 123-123 or 234234x2

If we only need commas and decimals, perhaps we could use: `^-?[0-9]+[,.]?[0-9]*$`

src/dapps/aave-dapp/components/AaveSummary.vue
```diff
@@ -156,6 +157,15 @@ export default {
...
   computed: {
+    actionTitle() {
+      if (this.isBorrow) {
+        return 'Borrow';
+      }
+      if (this.isRepay) {
+        return 'Repay';
+      }
+      return 'Deposit';
+    },
```
rsteubs, line 168: Do we need to include isCollateral in this / is isCollateral used anywhere?  Or will Deposit always be the default state.

gesseekur: we don't because its hidden if its collateral! but good call

src/dapps/aave-dapp/components/AaveAmountForm.vue
gesseekur: ooh got it

gesseekur: not sure if this regex works its erroring out for me but i am trying ^(0|[1-9][0-9]*)$

rsteubs: Yeah I think it's the linter.  Could surpress it or use this `^-?[0-9]+[.]?[0-9]*$` to single out the period character

gesseekur: okay ill try it!

gamalielhere (approved):

rsteubs (approved):

kvhnuke:
this requires aave v2 api

hsurf22 (approved):

PR #3194, opened 2021-07-01
v6/ui: add renames and text changes for body header etc

rsteubs:
### Feature

Updating texts according to V6 specifications:


package.json
```diff
@@ -11,7 +11,8 @@
     "build:history": "ROUTER_MODE=history npm run build",
     "test": "npm run update:lists && jest --clearCache && npm run test:unit",
     "codecov": "npm run test && codecov",
-    "dev": "npm run update && npm run update:mainlists && vue-cli-service serve --watch",
+    "juststart": "vue-cli-service serve --watch",
```
olgakup, line 14: whats this new thing?

package.json
```diff
@@ -11,7 +11,8 @@
...
+    "dev": "npm run update && npm run juststart",
```
olgakup, line 15: and not sure why this was changed?

src/core/components/TheWrapperDapp.vue
```diff
@@ -89,7 +89,7 @@ export default {
   methods: {
     onClose() {
       this.$router.push({
-        name: 'Dapps'
+        name: 'DApps'
```
olgakup, line 92: can you revert this back please? this is actual route name  and they all should be lower case

src/core/router/routes-wallet.js
```diff
@@ -44,7 +44,7 @@ export default {
       beforeEnter: swapRouterGuard
     },
     {
-      path: 'dapps',
+      path: 'DApps',
```
olgakup, line 47: can you revert this back please? this is actual route name and they all should be lower case

olgakup (changes requested):

package.json
rsteubs: [package.json](https://github.com/MyEtherWallet/MyEtherWallet/blob/v6/develop/package.json) in v6/develop

package.json
rsteubs: [package.json](https://github.com/MyEtherWallet/MyEtherWallet/blob/v6/develop/package.json) in v6/develop, just keeping them in sync

gamalielhere (approved):

PR #3198, opened 2021-07-01
v6/txt-records-and-multicoin-issue

gamalielhere:
### Devop

* \[x] Updated CHANGELOG.md
* \[x] Add PR label
* \[x] No hard-coded strings


src/dapps/ens-manager-dapp/components/manage/ManageUploadWebsite.vue
```diff
@@ -122,7 +126,12 @@ export default {
       this.uploadFile(e.target.files[0]);
     },
     setHash(val) {
-      this.ipfs = val;
+      this.error =
+        val === ''
```
rsteubs, line 130: Can val ever be anything other than '' ?  Would it be better to use == to account for other falsy values?

src/dapps/ens-manager-dapp/components/manage/ManageMulticoins.vue
```diff
@@ -39,45 +39,70 @@ export default {
...
-          this.errors[coin.symbol] = '';
+          this.errors[coin] = '';
         }
       }
+    },
+    submit() {
+      /** Creates a no reference clone to be submitted to the contract */
+      const copyMulticoin = Object.keys(this.multicoin).map(item => {
+        const newItem = Object.assign({}, _.clone(this.multicoin[item]));
+        newItem.value = this.setCoins[item];
```
kvhnuke, line 102: only add it if the old value changed, if not itll be an expensive tx since itll set whole bunch of 000...000 addresses

src/dapps/ens-manager-dapp/components/manage/ManageUploadWebsite.vue
kvhnuke: I think since the value is coming from text field it can only be '' but yea good observation, itll probably be better to use ==

kvhnuke (changes requested):

src/dapps/ens-manager-dapp/components/manage/ManageUploadWebsite.vue
gamalielhere: @rsteubs the input component returns null when cleared using the 'x' button within the component. I know @gesseekur changed this so this might now be necessary at all.

kvhnuke (approved):

gesseekur (approved):

PR #3224, opened 2021-07-06
v6/graphic-changes-ui

rsteubs:
### UI: 
- My-Tokens-Table svg
- Swap-Page png
- Send Transaction svg
- Replace google play and ios app store badges with svg
- Highlights and on cursor for wallet actions mobile drawer
- Border radius changes for all v dialogs


src/core/components/AppGetStarted.vue
```diff
@@ -6,20 +6,29 @@
         <div class="d-flex align-center px-6">
           <div>
             <h1 class="white--text mb-3">Ready to explore Ethereum?</h1>
-            <p class="white--text ma-0">
-              Still don't know what MEW is?
+            <div align="left" justify="space-around" class="mt-5">
```
gesseekur, line 9: wow had no idea you can do `align` like that on a div 

not sure if justify is doing anything though i can't find any docs about it, if you could link me that'd be great!

rsteubs: Oh yeah that can be removed, it's only for some Vue components.

gamalielhere (approved):

gesseekur (approved):

PR #3234, opened 2021-07-07
Bug/fix invalid values for swap amount

rsteubs:
### Bugs

- Adds invalid decimals point check on the amount in SWAP
   previously could enter any values.
- Fix 'Max' button causing negative values


src/modules/swap/ModuleSwap.vue
```diff
@@ -881,11 +883,13 @@ export default {
...
-      this.tokenInValue = this.isFromTokenMain
-        ? new BigNumber(this.availableBalance)
-            .minus(fromWei(MIN_GAS_WEI))
-            .toFixed()
-        : this.availableBalance.toFixed();
+      const availableBalanceMinusGas = new BigNumber(this.availableBalance)
+        .minus(fromWei(MIN_GAS_WEI))
+        .toFixed();
+      this.tokenInValue =
+        this.isFromTokenMain && availableBalanceMinusGas > 0
```
olgakup, line 890: So the only reason you want to use availableBalanceMinusGas is when you actually swap main chain currency Like ETH, BNB  and Matic (isFromTokenMain). You need to subtract min gas fee from the balance.
Here you are checking if the (isFromTokenMain and available is greater then 0, if they availableBalanceMinusGas > 0 --> then you get the whole available balance when you should be setting it to 0

olgakup (changes requested):

src/modules/swap/ModuleSwap.vue
kvhnuke: I think this is fine but `availableBalanceMinusGas > 0` you cant compare bignumbers with `> <` signs you have to use `.gt or .gte`

kvhnuke: oh @olgakup I see what you mean

gamalielhere (approved):

src/modules/swap/ModuleSwap.vue
```diff
@@ -883,13 +883,16 @@ export default {
...
+      let availableBalanceMinusGas = new BigNumber(this.availableBalance)
         .minus(fromWei(MIN_GAS_WEI))
         .toFixed();
-      this.tokenInValue =
-        this.isFromTokenMain && availableBalanceMinusGas > 0
-          ? availableBalanceMinusGas
-          : this.availableBalance.toFixed();
+
+      availableBalanceMinusGas =
+        availableBalanceMinusGas > '0' ? availableBalanceMinusGas : '0';
```
kvhnuke, line 822: @rsteubs you cant use > with bignumbers, it is currently just comparing strings

olgakup (approved):

kvhnuke (approved):

PR #3239, opened 2021-07-08
fix: 🐛 show private key in print

olgakup:
### Bug

Private key was not shown in paper wallet for keystore and mnemonic
* \[x] Updated CHANGELOG.md
* \[x] Add PR label


gamalielhere (approved):

src/modules/balance/components/PaperWalletToDisplay.vue
```diff
@@ -136,7 +136,7 @@ export default {
   computed: {
     ...mapState('wallet', ['address', 'instance', 'isHardware']),
     key() {
-      if (this.address && !this.showPrivateKey) {
+      if (this.address && this.showPrivateKey) {
```
kvhnuke, line 139: There is this.instance.isPubOnly if that is true no private key you can use that for this if statement

kvhnuke (changes requested):

kvhnuke (approved):

Badameron31:
This sucks

PR #3242, opened 2021-07-08
fix: 🐛 apollo pending error

gesseekur:
### Bug

* \[x] Updated CHANGELOG.md
* \[x] Add PR label


gamalielhere (approved):

src/core/mixins/handlerWallet.mixin.js
```diff
@@ -19,6 +19,9 @@ export default {
     getLatestPrices: {
       query: getLatestPrices,
       pollInterval: 600000,
+      skip() {
+        return !this.isEthNetwork;
```
kvhnuke, line 23: This is needed, no matter the network is, this is how we get the prices for the tokens so don't skip

kvhnuke (changes requested):

src/views/TheWalletView.vue
```diff
@@ -40,7 +40,7 @@ export default {
   computed: {
     ...mapState('wallet', ['address', 'web3', 'identifier']),
     ...mapState('global', ['online', 'gasPriceType', 'baseGasPrice']),
-    ...mapGetters('global', ['network']),
+    ...mapGetters('global', ['network', 'isEthNetwork']),
```
kvhnuke, line 43: I don't think this is needed either

kvhnuke (changes requested):

src/core/mixins/handlerWallet.mixin.js
gesseekur: removed

src/views/TheWalletView.vue
gesseekur: removed

kvhnuke (approved):

andrewdahn (approved):

PR #3269, opened 2021-07-13
fix: 🐛 reset is not a function

rsteubs:
### Bug
- [x] Updated CHANGELOG.md
- [x] Link to issue: [Link](https://app.asana.com/0/1163629870679883/1200594450451403/f)
- [x] Add PR label

RCA: Removing a leftover function call from a refactor.  Since vue data() is instantiated from a function, a method call to reset the data is not needed.


src/dapps/ens-manager-dapp/TheENSManagerLayout.vue
```diff
@@ -504,7 +504,6 @@ export default {
       this.loadingCommit = false;
       this.name = '';
       this.nameHandler = {};
-      this.$refs.registerDomain.reset();
```
kvhnuke, line 507: but dont we need to reset the field?

kvhnuke (changes requested):

src/dapps/ens-manager-dapp/TheENSManagerLayout.vue
rsteubs: I don't think so since as soon as you click close, the instantiation gets destroyed.  When you reopen, it just uses the default data in ModuleRegisterDomain.vue which starts them off at step 1

`data() {
    return {
      duration: 1,
      onStep: 1,`

kvhnuke (approved):

gamalielhere (approved):

PR #3276, opened 2021-07-14
fix: 🐛 handle reported toLowerCase issues in sentry

gamalielhere:
### Bug

* \[x] Updated CHANGELOG.md
* \[x] Add PR label


src/core/store/external/getters.js
```diff
@@ -92,42 +92,45 @@ const getCoinGeckoTokenById = state => cgid => {
...
  */
 const contractToToken =
-  (state, getters, rootState, rootGetters) => contractAdress => {
-    contractAdress = contractAdress.toLowerCase();
-    let tokenId = platformList[contractAdress];
-    if (contractAdress === MAIN_TOKEN_ADDRESS) {
-      tokenId = rootGetters['global/network'].type.coingeckoID;
-      const networkType = rootGetters['global/network'].type;
+  (state, getters, rootState, rootGetters) => contractAddress => {
+    if (contractAddress) {
```
kvhnuke, line 96: can you actually figure out where it calls this without a contract address? lets find the root cause

src/modules/address-book/components/AddressBookAddEdit.vue
```diff
@@ -212,7 +212,7 @@ export default {
       }
     },
     setAddress(value) {
-      this.addressToAdd = value;
+      this.addressToAdd = value ? value : '';
```
kvhnuke, line 215: is it going to add  '' as the address?

src/modules/swap/ModuleSwap.vue
```diff
@@ -905,6 +905,15 @@ export default {
...
     setTokenInValue: _.debounce(function (value) {
+      /**
+       * Ensure that both pairs have been set
+       * before calling the providers
+       */
+      const hasPair =
+        this.toTokenType.symbol &&
+        this.fromTokenType.symbol &&
+        !_.isEmpty(this.toTokenType) &&
+        !_.isEmpty(this.fromToTokenType);
```
kvhnuke, line 916: I think these two are redundant since if(this.toTokenType.symbol) tjem !_.isEmpty(this.toTokenType) is always true

kvhnuke (changes requested):

src/modules/address-book/components/AddressBookAddEdit.vue
gamalielhere: only if value is falsy which cascades to the lowercase error since the address isn't a string anymore

src/modules/swap/ModuleSwap.vue
gamalielhere: not necessarily, toTokenType is technically not empty when it's "Select Token" but is also not a valid to token

kvhnuke (approved):

gesseekur (approved):

PR #3298, opened 2021-07-19
fix: 🐛 sentry issue 2507823347

gesseekur:
### Bug

* \[x] Updated CHANGELOG.md
* \[x] Add PR label

https://app.asana.com/0/1163629870679883/1200594450451358


src/modules/access-wallet/hybrid/handlers/MEWconnect/index.js
```diff
@@ -29,9 +29,11 @@ class MEWconnectWallet {
     this.mewConnect = await connection.createWalletOnly(
       store.getters['global/network']
     );
-    this.mewConnect.disconnect = () => {
-      this.mewConnect.mewConnect.disconnectRTC();
-    };
+    if (this.mewConnect) {
```
rsteubs, line 32: I would figure that `await` would prevent this line from happening, is this check for when connection.createWalletOnly is returning null?

gesseekur: yep im not 100% sure where the sentry error is coming from - so I just added the check incase but youre right this one shouldn't error out. I'll remove it. I think It is most likely coming from wallet link

src/modules/access-wallet/hybrid/handlers/WalletConnect/index.js
```diff
@@ -34,11 +34,12 @@ class WalletConnectWallet {
         store.dispatch('wallet/removeWallet');
       }
     });
-
-    this.walletConnect.disconnect = () => {
-      this.isKilled = true;
-      this.walletConnect.killSession();
-    };
+    if (this.walletConnect) {
+      this.walletConnect.disconnect = () => {
```
kvhnuke, line 38: hmmm the error is `Cannot read property 'disconnect' of null` over here we are setting `.disconnect` property so this is not where `.disconnect()` gets called

kvhnuke: also the wallettype is web3wallet I'd do a search for `.disconnect()` and check those instances

kvhnuke (changes requested):

src/modules/access-wallet/hybrid/handlers/WalletConnect/index.js
gesseekur: ooohhhhh

gesseekur: fixed!

rsteubs (approved):

kvhnuke (approved):

PR #3340, opened 2021-07-26
bitbox01: hotfix on v5: make BitBox01 work with WebHID, drop U2F-hijack

benma:
BitBox01 is currently not functional on v6.0.3 and it does not seem
trivial to make it work.

The previous version is hosted at https://v5.myetherwallet.com/, but
there, BitBox01 support also broken.

This patch is a hot-fix for the v5 deployment that makes the BitBox01
integration functional again, using WebHID (Chrome only).

This will help BitBox01 users migrate their coins to a new wallet.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4874,19 +4874,6 @@
             "readable-stream": "~1.0.26"
           }
         },
-        "handlebars": {
-          "version": "4.7.7",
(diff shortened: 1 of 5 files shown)


benma:
It would be really awesome if you could merge this and deploy it to https://v5.myetherwallet.com/ :pray:

gamalielhere:
Moved to #3358

PR #3394, opened 2021-08-18
feat: update transaction fee

hsurf22:
### Feature

* \[x] Add PR label
   - eip1559 transaction fee interface

### DIFF
--- a/src/core/components/AppSimpleDialog.vue
+++ b/src/core/components/AppSimpleDialog.vue
@@ -1,0 +1,75 @@
+<template>
+  <!--
+  ==================================================================
+  Use (:value=""), (@close="")
+  or (v-model="") to control dialog
+  ==================================================================
+  -->
+  <v-dialog v-model="isDialogOpen" :width="width">
+    <div class="white pa-8">
+      <div class="d-flex align-start justify-space-between mb-7">
+        <div class="mew-heading-3 pr-5">
+          {{ title }}
+        </div>
+        <v-btn
+          color="textSecondary"
+          icon
+          class="mt-n2 mr-n2"
+          @click="closeDialog"
+        >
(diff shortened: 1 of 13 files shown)


src/core/components/AppSimpleDialog.vue
```diff
@@ -0,0 +1,75 @@
+<template>
```
gesseekur, line 1: I believe theres already a dialog component

src/core/components/AppTransactionFee.vue
```diff
@@ -0,0 +1,270 @@
+<template>
```
gesseekur, line 1: since this is only used at one place (swap) i think ..it can go under that folder instead of this one, this folder will eventually be removed

gamalielhere: This is fine since it'll be used on Send as well.

gesseekur: should it be moved to Core components then?

gamalielhere: Actually you're right, this is only used on Swap it looks like. It should just be placed within the swap folder

gamalielhere (changes requested):
Changes doesn't reflect current design, please check asana board for current design implementation

src/core/components/AppSimpleDialog.vue
hsurf22: I can't find popup or dialog component like this anywhere.

src/core/components/AppTransactionFee.vue
hsurf22: We use this in send page too.

hsurf22:
> Changes doesn't reflect current design, please check asana board for current design implementation

I fixed some UIs.

PR #3414, opened 2021-08-25
devop: 🔧 fix swap icons and hide provider list when amount error

gamalielhere:
### Bug

* \[x] Add PR label


src/modules/swap/ModuleSwap.vue
```diff
@@ -782,6 +782,9 @@ export default {
     },
     mainTokenDetails() {
       this.setDefaults();
+    },
+    amountErrorMessage(newVal) {
+      if (newVal !== '') this.availableQuotes.splice(0);
```
rsteubs, line 787: Will it always be the 0th value to splice?

gesseekur (approved):

rsteubs (approved):

andrewdahn (approved):

src/modules/swap/ModuleSwap.vue
gamalielhere: yeah, it's resetting the array back to empty

PR #3416, opened 2021-08-26
feat: ✨  Added Gochain network (ws API)

r-gochain:
### Feature

* \[ ] Added entry to ./changelog
* \[ ] Is this a user submitted feature request?
  * \[ ] Link to issue:
* \[ ] Add test cases or procedure
* \[ ] Add PR label
* Adding Gochain websocket API endpoints.


src/utils/networks/types/GO.js
```diff
@@ -0,0 +1,25 @@
...
+  currencyName: 'GO',
+  isTestNetwork: false,
+  isEthVMSupported: {
+    supported: false,
+    url: null,
+    blockExplorerTX: '',
+    blockExplorerAddr: '',
+    websocket: null
+  },
+  gasPriceMultiplier: 1
```
kvhnuke, line 24: can you please add the `coingeckID` I believe it is 'gochain' 
https://api.coingecko.com/api/v3/coins/list?include_platform=true

kvhnuke (changes requested):

src/utils/networks/types/GO.js
r-gochain: Added (didn't find any other usages of coingeckID in the code)

rsteubs (approved):

gamalielhere:
Added into #3442

gamalielhere:
@r-gochain we're running into an issue with the node you submitted: "received a broken close frame containing a reserved status code.". Please take a look at it as we might have to remove gochain until this is resolved.

treeder:
Hi, we'll look into this right away. Could you could leave it in for another day or so so we can debug?

treeder:
@gamalielhere do you know how to reproduce the issue?  I just tried sending some GO using metamask via MyEtherWallet with no problem.

r-gochain:
@gamalielhere could you please share the way to reproduce this problem? I just checked and I can connect using different wallets and wscat from the command line without any issues. Maybe some code to reproduce?

image

gamalielhere:
Screen Shot 2022-01-18 at 2 55 46 PM

@r-gochain @treeder this is on MEW after accessing the wallet or any type of web3 calls with the gochain node it seems like

r-gochain:
Thanks for sharing, I'll reproduce and try to fix.

r-gochain:
@gamalielhere could you please check, we've updated the proxy and now it doesn't throw such errors

gamalielhere:
@r-gochain seems like the error is gone now, but for some reason it's still not resolving the response to web3. We're also looking into it on our end.

r-gochain:
@gamalielhere I can see that there is communication going through, but sometimes is fails to reuse the opened websocket (sometimes the socket is already closed). Anything I could do from my end to help you?

image

r-gochain:
@gamalielhere please test once again, I've submitted a couple more fixes, now it shouldn't occasionally break the connection

gamalielhere:
@r-gochain just got done testing it, the new changes seems to have fixed the issue we're having. Thanks for looking into it and fixing it!

PR #3450, opened 2021-09-13
fix: 🐛  expected public key to be an Uint8Array with length

rsteubs:
RCA:  Web3 / MetaMask sends back an accountsChanged event along with an undefined account object when you lock your wallet first time.  Undefined account => undefined public key.

### Bug
* \[x] Add PR label


andrewdahn (approved):

src/views/TheWalletView.vue
```diff
@@ -78,7 +78,12 @@ export default {
     this.web3.eth.clearSubscriptions();
   },
   methods: {
-    ...mapActions('wallet', ['setBlockNumber', 'setTokens', 'setWallet']),
+    ...mapActions('wallet', [
+      'setBlockNumber',
+      'setTokens',
+      'setWallet',
+      'removeWallet'
```
gamalielhere, line 85: doesn't seem like this is being used.

gamalielhere (changes requested):

src/views/TheWalletView.vue
rsteubs: Oh yeah, thanks.  The idea was if metamask was sending an empty account array every time the user locked metamask, we could use that to just log them out.  However, looks like the empty account / accountsChanged event is only sent on first time lock

I'll remove this change.

gamalielhere (approved):

PR #3531, opened 2021-11-02
fix: 🐛 module swap then undefined

andrewdahn:
### Bug

* \[x] Add PR label


rsteubs (approved):

src/modules/swap/handlers/handlerSwap.js
```diff
@@ -83,6 +83,7 @@ class Swap {
   getTrade(tradeInfo) {
     for (const p of this.providers) {
       if (p.provider === tradeInfo.provider) return p.getTrade(tradeInfo);
+      throw new Error('This provider is not available.');
```
gamalielhere, line 86: This throws since get trade is also used to fetch for quotes. Please check Asana card for the issue explanation or let me know if you have questions.

gamalielhere (changes requested):

src/modules/swap/handlers/handlerSwap.js
andrewdahn: @gamalielhere Applied try catch as per Asana explanation

hsurf22 (approved):

gamalielhere (approved):

PR #3567, opened 2021-11-18
fix: 🐛 fix mew wallet banner invalid url bug

hsurf22:
### Fix

* \[x] Add PR label
- https://app.asana.com/0/1125971615701970/1201383357900967
- https://sentry.io/organizations/myetherwallet-inc/issues/2777631614/?project=5652727&query=is%3Aunresolved&sort=freq&statsPeriod=14d

Screenshot_20211118-152252_Chrome


src/views/components-default/TheDefaultWalletBanner.vue
```diff
@@ -80,9 +80,7 @@
             class="mew-body font-weight-medium footer-text"
           >
             <p>{{ $t('home.mew-wallet-modal.footer-text') }}</p>
-            <a
-              :href="isMobile() ? btn.url : 'https://mewwallet.com'"
-              target="_blank"
+            <a href="https://mewwallet.com" target="_blank"
```
rsteubs, line 83: I think we still want to keep the mobile links, this link (mewwallet.com) is for desktop.

hsurf22: This code is outside of the v-for loop,so the (btn) doesn't exist at this point.
I think the mobile and desktop link for this should be the same???
Please see the picture.

gamalielhere (approved):

rsteubs (approved):

PR #3632, opened 2022-01-20
fix: 🐛 Added Mount Check

jmjles:
### Devop

* \[x] Add PR label


src/modules/swap/ModuleSwap.vue
```diff
@@ -901,8 +903,10 @@ export default {
       this.addressValue = {};
       this.selectedProvider = {};
       this.localGasPrice = '0';
-      this.$refs.toToken.clear();
-      this.$refs.amountInput.clear();
+      if (this.mounted) {
```
gamalielhere, line 906: no need for the mounted variable. just check for whether they're not null when calling `this.clear`. before calling them.

gamalielhere (changes requested):

gamalielhere (approved):

andrewdahn (approved):

PR #3641, opened 2022-01-25
Added MintMe websocket support 

Denlev:
Mintme network is not listed in the new MEW version 6, yet it's there on version 5. This PR contains ws configuration for the MINTME network.


gamalielhere:
can you check the eth_getTransactionCount? the calls doesn't seem to be resolving back to web3. We had a similar issue with GoChain here: #3416

Issue #3618, opened 2022-01-12
Please bump bitbox02 package to v0.12.0 to fix WebHID access on Chrome


benma:
@gamalielhere could you please bump this to 0.12.0: https://github.com/MyEtherWallet/MyEtherWallet/blob/91780bed6f9603b4200990c4ecb60bdbf22971a1/package.json#L67

WebHID access to the BitBox02 stopped working as of Chrome version
97 due to incorrect vendor/product filter names. See also:
digitalbitbox/bitbox02-api-js#62

Simply upgrading the package solves the problem.

Would be great to have this deployed alongside https://github.com/MyEtherWallet/MyEtherWallet/pull/3611 soon, so BitBox02 users can use v6 without a hitch again.

Let me know if I can be of any assistance.

Thank you.

gamalielhere:
@benma are there changes we need to do on our end or just the package update?

benma:
@gamalielhere just updating is enough.

benma:
Great, thanks!

PR #3643, opened 2022-01-26
bitbox01: fix WebHID filters to restore access in Chrome 97+

benma:
WebHID access to the BitBox01 stopped working as of Chrome version
97 due to incorrect vendor/product filter names. See also:
digitalbitbox/bitbox02-api-js#62

The same issue was fixed for the BitBox02 previously here:
https://github.com/MyEtherWallet/MyEtherWallet/issues/3618


benma:
@gamalielhere would be great if you could merge and deploy this to v5 so bitbox01 users regain access.

gamalielhere:
@benma do we upgrade the bitbox02 package for this version or do we just ignore that for now?

PR #3686, opened 2022-02-17
Feat: Currency Localization

jmjles:
### Devop

* \[x] Add PR label


gamalielhere (changes requested):
can you move this into a getter? We have 17 instances of the same code over and over.

gamalielhere (approved):

gamalielhere (approved):

PR #3729, opened 2022-03-15
feature: ✨ intercom-set

NewCodeOnDaBlock:
### Feature

 * \[x] Link to issue:
* \[x] Add PR label


src/main.js
```diff
@@ -61,9 +61,14 @@ new Vue({
   apolloProvider,
   vuetify,
   beforeCreate() {
-    this.$intercom.boot({
-      user_id: uuidv4()
-    });
+    const userId = this.$route.query.intercomid
```
gamalielhere, line 64: I would check whether intercomid exists here as well.

src/main.js
```diff
@@ -61,9 +61,14 @@ new Vue({
...
+      ? this.$route.query.intercomid
+      : uuidv4();
+    if (userId) {
```
gamalielhere, line 67: won't need to check this since if intercomid is not there, it will be filled with the uuidv4() function

gamalielhere (changes requested):

src/main.js
```diff
@@ -61,9 +61,13 @@ new Vue({
   apolloProvider,
   vuetify,
   beforeCreate() {
-    this.$intercom.boot({
-      user_id: uuidv4()
-    });
+    const userId = this.$route.query.intercomid
+      ? this.$route.query.intercomid
+      : uuidv4();
+    !this.$route.query.intercomid
```
gamalielhere, line 67: This is unnecessary as you already checked it the line before. Just call boot with userId.

gamalielhere (changes requested):

src/main.js
NewCodeOnDaBlock: fixed

gamalielhere (approved):

andrewdahn (approved):

hsurf22 (approved):

PR #3740, opened 2022-03-22
Sentry: Request String MewAPI

jmjles:
### Devop

* \[x] Add PR label


src/main/sentry.js
```diff
@@ -29,6 +31,11 @@ Sentry.init({
       service: service,
       walletType: identifier
     };
+    const err = event.exception.values[0].value;
+    if (knownErrors[err]) {
```
gamalielhere, line 35: Use the same check for `WalletErrorHandler.js` so it doesn't have to be 1:1 match for these errors

gamalielhere (changes requested):

andrewdahn (approved):

PR #3829, opened 2022-05-13
Fix: sentry expected private key 32 length

Abyscuit:
### Fix

  * \[ ] Link to issue: https://sentry.io/organizations/myetherwallet-inc/issues/2786738543/?project=5652727&query=is%3Aunresolved
* \[x] Add PR label


jmjles (approved):

src/modules/access-wallet/common/WalletInterface.js
```diff
@@ -28,10 +28,22 @@ class WalletInterface {
...
+          if (!isValidPrivate(_privKey))
+            throw new Error(
+              'Private key does not satisfy the curve requirements (ie. it is invalid)'
+            );
+        } catch (e) {
+          throw new Error(
+            'Private key does not satisfy the curve requirements (ie. it is invalid)'
+          );
+        }
+      }
```
kvhnuke, line 46: you can catch both in one if
```
f (_privKey.length !== 32 || !isValidPrivate(_privKey) ) {
        throw new Error(
          'Private key does not satisfy the curve requirements (ie. it is invalid)'
        );
      }
```

kvhnuke (dismissed):

andrewdahn (approved):

PR #3832, opened 2022-05-16
feat: ✨ implementation of eip-1271 signatures support

0xdex18:
This PR implements EIP-1271 signatures verifications modifying the UI and using the [Ambire signature-validator package](https://www.npmjs.com/package/@ambire/signature-validator)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
   "dependencies": {
     "@aave/protocol-js": "1.0.3",
     "@actions/github": "5.0.1",
+    "@ambire/signature-validator": "^1.0.3",
     "@coolwallets/eth": "0.2.0-beta.1",
     "@coolwallets/transport-web-ble": "0.1.0",
     "@coolwallets/wallet": "0.2.0-beta.1",
--- a/src/modules/message/ModuleMessageVerify.vue
+++ b/src/modules/message/ModuleMessageVerify.vue
@@ -64,15 +64,21 @@ export default {
   },
   computed: {},
   mounted() {
-    this.signAndVerify = new SignAndVerifyMessage(this.web3, this.address);
+    this.signAndVerify = new SignAndVerifyMessage(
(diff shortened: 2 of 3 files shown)


kvhnuke:
we prefer not to import libraries that aren't verified/popular can you show us a use case where our signature verification fail. So we can make a separate PR to fix it

0xdex18:
In all cases where the signer is a smart contract wallet, like Gnosis Safe, Ambire, Argent, the normal ecrecover-based signature recovery simply does not work. The correct way to verify a signature by any UI, is to validate it with a transactionless call to the contract to isValidSignature (eip-1271)

This issue of not supporting 1271 signatures is not only affecting Ambire Wallet, but all the Smart Wallets (Gnosis safe, Argent, etc...) as they use smart contract to verify the validity of their signatures.

@ambire/signature-validator is a very simple yet efficient lib that covers all the signatures verifications, (regular, 712 and 1271) for this purpose. The lib was created especially because there were not such a lib to cover signature verification in a simple way.

The library is only ~60 lines of code: [@ambire/signature-validator](https://github.com/AmbireTech/signature-validator), so you can include it directly to minimize the potential security impact - and it only has one dependency, which is ethers which you already use

Example of 1271 signature that IS 1271 valid in MEW (polygon network): 

{
  "address": "0x4836a472ab1dd406ecb8d0f933a985541ee3921f",
  "msg": "0x416d626972652077616c6c65742074657374206d657373616765",
  "sig": "e0b6dcb5f71d3cb6978aa144fa03cfd717920690cc7607ccb489e1a4de7a0bcc622d11ebcbbf3afa936bf7f5089c3e76c2aee0d0d4be3d1465fe57a73bd3c1be1b00",
  "version": "3",
  "signer": "MEW"
}

0xdex18:
+1

PR #3873, opened 2022-05-31
fix: 🐛null reading error handler

NewCodeOnDaBlock:
### Fix

* \[x] Add PR label


jmjles (approved):

src/modules/send/ModuleSend.vue
```diff
@@ -700,7 +700,7 @@ export default {
         })
         .catch(error => {
           this.clear();
-          this.instance.errorHandler(error);
+          this.instance?.errorHandler(error);
```
gamalielhere, line 703: check if instance exists, if not, use Toast handler.

PR #3875, opened 2022-06-01
fix: sentry cannot read null 'getAddressString'

Abyscuit:
### Fix

* \[x] Add PR label


src/utils/web3-provider/methods/eth_accounts.js
```diff
@@ -1,8 +1,13 @@
 import { toPayload } from '../jsonrpc';
 export default async ({ payload, store }, res, next) => {
   if (payload.method !== 'eth_accounts') return next();
-  res(
-    null,
-    toPayload(payload.id, [store.state.wallet.instance.getAddressString()])
-  );
+  if (!store.state.wallet) return next();
```
gamalielhere, line 4: might need to just to res with an error here since it shouldn't happen.

src/utils/web3-provider/methods/eth_accounts.js
```diff
@@ -1,8 +1,13 @@
...
+  try {
+    res(
+      null,
+      toPayload(payload.id, [store.state.wallet.instance.getAddressString()])
+    );
+  } catch (err) {
+    return next(err);
```
gamalielhere, line 11: use `res` for the error here.

gamalielhere (changes requested):

gamalielhere (approved):

PR #3918, opened 2022-06-13
Fix: sentry json response has no error or result

Abyscuit:
### Fix

* \[x] Add PR label


package-test.js
```diff
@@ -2,8 +2,8 @@
 // to date. If dependencies are too outdated, exit with an error, failing `npm
 // run update:packages` and thus eventually the entire build.
 
-const package = require('./package.json');
-const packageJson = require('package-json');
+const package = import('./package.json');
```
gamalielhere, line 5: import wont work here as import is not supported by node natively.

gamalielhere (changes requested):

gamalielhere (approved):

andrewdahn (approved):

PR #3928, opened 2022-06-15
fix: signer issue with cool wallet swap

andrewdahn:
### Fix

* \[x] Add PR label


src/modules/confirmation/ModuleConfirmation.vue
```diff
@@ -161,7 +163,9 @@
                     />
                     <v-icon
                       v-show="
-                        !isBatch && Object.keys(signedTxObject).length !== 0
+                        !isBatch &&
```
gamalielhere, line 166: can you move these into computed props so its cleaner

gamalielhere (approved):

PR #3946, opened 2022-06-21
Sentry: Cannot read properties of undefined (reading 'message')

Abyscuit:
### Fix

* \[x] Add PR label


src/views/layouts-default/TheAccessWalletLayout.vue
```diff
@@ -324,7 +324,7 @@ export default {
             this.$router.push({ name: ROUTES_WALLET.WALLETS.NAME });
           }
         } catch (e) {
-          Toast(e.message, {}, WARNING);
+          Toast(e.hasOwnProperty('message') ? e.message : e, {}, WARNING);
```
gamalielhere, line 327: just pass e in this case. it should be handled within the Toast module.

gamalielhere (changes requested):

NewCodeOnDaBlock (approved):

PR #3979, opened 2022-07-01
Fix: address book validation

Abyscuit:
### Fix

* \[x] Add PR label


gamalielhere (approved):
Also move Ethereum and BTC on the list so it resolves faster for those.

hsurf22 (approved):

PR #3982, opened 2022-07-05
Fix: max button when balance zero

Abyscuit:
### Fix

* \[x] Add PR label


src/modules/swap/ModuleSwap.vue
```diff
@@ -440,7 +440,8 @@ export default {
      * if native token, return empty
      */
     maxBtn() {
-      return this.isFromNonChain
+      return this.isFromNonChain ||
+        this.amountErrorMessage === this.errorMsgs.doNotOwnToken
```
gamalielhere, line 444: you can use `this.availableBalance` to check if the balance is 0

gamalielhere (changes requested):

gamalielhere (approved):

hsurf22 (approved):

PR #4012, opened 2022-07-16
Fix: swap error 404

Abyscuit:
### Fix

* \[x] Add PR label


src/modules/swap/ModuleSwap.vue
```diff
@@ -1347,6 +1347,16 @@ export default {
...
       if (trade instanceof Promise) {
         trade.then(tradeResponse => {
+          if (!tradeResponse) {
+            const index = this.availableQuotes.indexOf(swapObj.quote);
+            if (index > -1) {
+              // Remove the quote
+              this.availableQuotes.splice(index, 1);
+            }
+            // Hide the 404 & 500 Error Toast
+            Toast('There was an issue with the provider', {}, ERROR);
```
gamalielhere, line 1357: use 'this.feeError' so it shows in the ui and not just in the Toast.

gamalielhere (changes requested):

gamalielhere (approved):

PR #4036, opened 2022-07-25
devop: update mobile tools page

hsurf22:
### Devop

* \[x] Add PR label
   - update mew-tabs and mew-stepper on tabs page
Screenshot from 2022-08-01 19-48-35

Screenshot from 2022-07-25 15-26-43


src/core/plugins/vuetify.js
```diff
@@ -86,6 +86,7 @@ import Vuetify, {
   VScrollYTransition,
   VSkeletonLoader,
   VChip,
+  VChipGroup,
```
gamalielhere, line 89: Is this necessary? doesn't seem like it's being used anywhere.

gamalielhere (changes requested):

src/core/plugins/vuetify.js
hsurf22: Yeah....
That is for the mobile friendly tab buttons.
Screenshot from 2022-07-28 21-04-01

gamalielhere (approved):

src/core/plugins/vuetify.js
hsurf22: I already added VChipGroup to mew-components, but I still get an VChipGroup is missing error. After I add VChipGroup locally, it doesn't show the error.
https://github.com/MyEtherWallet/mew-components/blob/master/src/plugins/vuetify.js

Screenshot from 2022-08-05 15-10-01

NewCodeOnDaBlock (approved):

PR #4063, opened 2022-08-01
Fix: Moonpay sell dollar values

Abyscuit:
### Fix

* \[x] Add PR label


src/modules/moon-pay/components/MoonPaySellComponent.vue
```diff
@@ -421,6 +423,7 @@ export default {
     this.locGasPrice = this.gasPriceByType(this.gasPriceType);
   },
   methods: {
+    ...mapActions('external', ['setCoinGeckoTokens']),
```
gamalielhere, line 426: Where is this being used?

gamalielhere (changes requested):

src/modules/moon-pay/components/MoonPaySellComponent.vue
Abyscuit: It is being called in the walletHandler mixin, when I remove it I get an error.
image

andrewdahn (approved):

PR #4087, opened 2022-08-05
Fix: error switching networks metamask

NewCodeOnDaBlock:
### Fix

* \[x] Add PR label


gamalielhere (approved):

andrewdahn (approved):

src/modules/swap/ModuleSwap.vue
```diff
@@ -1108,12 +1108,6 @@ export default {
       const findToken = this.actualFromTokens.find(item => {
         if (item.contract === this.defaults.fromToken) return item;
       });
-      if (
```
gamalielhere, line 1111: why was this removed? this doesn't seem to be related to the issue.

gamalielhere (changes requested):

src/modules/swap/ModuleSwap.vue
NewCodeOnDaBlock: I've added it back However, I've pinpointed the issue to be here in this block of code. I've been trying to understand why the return on the 'if' on line 1111 returns mainTokenDetails based on the check. So for now, i've set it to 'findToken' and seems to fix the issue.

src/modules/swap/ModuleSwap.vue
```diff
@@ -1123,7 +1123,7 @@ export default {
         this.defaults.fromToken === MAIN_TOKEN_ADDRESS &&
         new BigNumber(this.balanceInETH).gt(0)
       ) {
-        return this.mainTokenDetails;
+        return findToken;
```
gamalielhere, line 1126: why is this changed? was this relevant to the fix?

gamalielhere (changes requested):

gamalielhere (approved):

PR #4131, opened 2022-08-23
Fix: multiple ETH swap options

Abyscuit:
### Fix

  * \[ ] Link to issue:
* \[x] Add PR label


src/modules/swap/ModuleSwap.vue
```diff
@@ -693,9 +696,13 @@ export default {
...
      */
     fromTokens() {
-      return this.availableTokens.fromTokens.map(token => {
-        return localContractToToken[token.contract];
-      });
+      return this.availableTokens.fromTokens
+        .map(token => {
+          return localContractToToken[token.contract];
+        })
+        .filter(
```
gamalielhere, line 703: This filter should be in the initial setup tokens since this only needs to happen once.

gamalielhere (changes requested):

gamalielhere (approved):

NewCodeOnDaBlock (approved):

PR #4171, opened 2022-09-07
Fix: invalid return values when accessing new network on MetaMask

Abyscuit:
### Fix

  * \[ ] Link to issue:
* \[x] Add PR label


gamalielhere (approved):

src/modules/network/components/NetworkSwitch.vue
```diff
@@ -323,7 +323,8 @@ export default {
             this.networkLoading = false;
             if (!this.isOffline) {
               const provider =
-                this.identifier === WALLET_TYPES.WEB3_WALLET
+                this.identifier === WALLET_TYPES.WEB3_WALLET &&
+                window.ethereum.isMetaMask
```
kvhnuke, line 327: this shouldn't be there as we accept extensions other than `metamask`

src/utils/web3-provider/providers/given-provider.js
```diff
@@ -52,6 +52,8 @@ class CustomRequestManager extends Web3RequestManager {
 class GivenProvider {
   constructor(host) {
     this.givenProvider = host;
+    if (typeof this.givenProvider === 'string')
+      throw new Error('Not supported network type');
```
kvhnuke, line 56: yea I dont think this is correct

kvhnuke: is there a way to replicate the issue?

src/views/TheWalletView.vue
```diff
@@ -205,11 +226,9 @@ export default {
         window.ethereum &&
         this.instance.identifier === WALLET_TYPES.WEB3_WALLET
       ) {
-        const networkId = await window.ethereum?.request({
-          method: 'net_version'
-        });
+        const chainId = window.ethereum.chainId;
```
kvhnuke, line 229: `window.ethereum.chainId` not always available and deprecated. use 
```
const chainId = await window.ethereum.request({
          method: 'eth_chainId'
        });
```

kvhnuke (changes requested):

src/utils/web3-provider/providers/given-provider.js
Abyscuit: I usually get this issue when I access my wallet with web3 and have the tab idle
GivenProviderString

hsurf22 (approved):

PR #4183, opened 2022-09-13
fix: 🐛 missing dollar sign on swap

NewCodeOnDaBlock:
### Fix

* \[x] Add PR label


src/modules/swap/ModuleSwap.vue
```diff
@@ -1293,9 +1294,8 @@ export default {
             item?.contract &&
             item?.contract?.toLowerCase() === value?.contract?.toLowerCase()
           )
-            return item;
+            value = foundToken[0];
```
Abyscuit, line 1297: You need to `return item` here then outside of the filter call set `value = foundToken[0];`

Abyscuit (changes requested):

Abyscuit (approved):

hsurf22 (approved):

PR #4190, opened 2022-09-14
fix: unavailable partner links

hsurf22:
### Fix

* \[x] Add PR label
   - unavailable partner link fixed


src/core/components/AppPartnersBlock.vue
```diff
@@ -41,8 +42,7 @@ export default {
         link: 'https://shiftcrypto.ch/?ref=mew'
       },
       {
-        img: require('@/assets/images/partners/secalot.png'),
-        link: 'https://www.secalot.com/'
+        img: require('@/assets/images/partners/secalot.png')
```
gamalielhere, line 45: Can we just remove it altogether?

gamalielhere (changes requested):

src/core/components/AppPartnersBlock.vue
hsurf22: removed!

andrewdahn (approved):

PR #4237, opened 2022-09-29
fix: 🐛 sentry type error balance card

NewCodeOnDaBlock:
### Fix

* \[x] Add PR label


src/modules/balance/ModuleBalanceCard.vue
```diff
@@ -308,7 +308,7 @@ export default {
      * returns @String
      */
     verifyAddressBody() {
-      return `To verify, check the address on your ${this.instance.meta.name} device & make sure it is the same address as the one shown below.`;
+      return `To verify, check the address on your ${this.instance?.meta.name} device & make sure it is the same address as the one shown below.`;
```
gamalielhere, line 311: move the check to a computed property and return an empty string if instance is null.

gamalielhere (changes requested):

package.json
```diff
@@ -158,7 +158,7 @@
     "postcss-url": "8.0.0",
     "prettier": "2.7.1",
     "print-js": "1.6.0",
-    "protobufjs": "7.1.1",
+    "protobufjs": "7.1.2",
```
gamalielhere, line 161: remove the second one.

gamalielhere (changes requested):

gamalielhere (changes requested):

Abyscuit (approved):

gamalielhere (approved):

PR #4541, opened 2023-03-01
fix: 🐛 return to checksum address value

ahsan-javaiid:
### Fix

* \[ ] Add PR label

### Description
The `return` statement is missing on `toChecksumAdd` call due to which address resolution is not working. Just fixed this bug by adding `return`. 

Very small pull request. Need to merge @gamalielhere

### DIFF
--- a/changelog/fix-4541.md
+++ b/changelog/fix-4541.md
@@ -1,0 +1,1 @@
+fix to checksum return value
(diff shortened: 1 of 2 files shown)


ahsan-javaiid:
cc: @gamalielhere

ahsan-javaiid:
Would be great if it could be included in [release](url) @gamalielhere 🙏

gamalielhere (approved):

PR #4587, opened 2023-04-20
FIX/RNS DAPP: Use web3 provider to derive signer in RNS manager dapp

ahsan-javaiid:
### Fix

* \[ ] Add PR label

### Description
This PR fixes web3 provider in rns manager dapp. Previously it deriving signer from priv key which is not available in all cases and also not a good approach. Now using web3 provider to derive signer and passing it to rns manager. 

### Testing
- Tested by connecting with metamask and sending / approving transactions
- Tested with mnemonic / seed phrase

### DIFF
--- a/changelog/fix-4587.md
+++ b/changelog/fix-4587.md
@@ -1,0 +1,1 @@
+fix web3 provider to derive signer in rns manager dapp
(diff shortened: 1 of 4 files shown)


ahsan-javaiid:
cc: @gamalielhere @alepc253

gamalielhere:
moved to #4588

PR #4674, opened 2023-07-18
fix: metamask infinite loop when network is not supported by metamask after mismatch

gamalielhere:
### Fix

* \[x] Add PR label


sgilflores3 (commented):
Thanks for all the help from all

PR #4993, opened 2024-08-28
FIX checksum error when you input a `Rootstock` checksummed address on send screen

ahsan-javaiid:
### Fix

  * \[ ] Link to issue:
* \[ ] Add PR label

### Description

This pr fixes `Error bad address checksum` thrown by ethers.js dependency used by ens.js dependency. Actually ethers.js rejects rootstock checksummed addresses so address needs to be lowercased before passing this function. 

### How to reproduce
- Select Rootstock network
- On send screen put a Rootstock checksummed address 

https://github.com/user-attachments/assets/13923f20-d92a-4120-a658-e5da4770e9c0

### Snapshot

Screenshot 2024-08-28 at 3 03 44 PM

### DIFF
--- a/src/modules/address-book/ModuleAddressBook.vue
+++ b/src/modules/address-book/ModuleAddressBook.vue
@@ -44,6 +44,7 @@ import WAValidator from 'multicoin-address-validator';
 import { isAddress, toChecksumAddress } from '@/core/helpers/addressUtils';
 import Resolver from '@/modules/name-resolver/index';
 import { ERROR, Toast } from '../toast/handler/handlerToast';
+import { ROOTSTOCK } from '@/utils/networks/types';
 
 const USER_INPUT_TYPES = {
   typed: 'TYPED',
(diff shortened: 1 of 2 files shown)


src/modules/address-book/ModuleAddressBook.vue
```diff
@@ -334,8 +334,13 @@ export default {
     resolveAddress: throttle(async function () {
       if (this.nameResolver) {
         try {
+          // Ethers.js rejects Rootstock checksummed address so use lowercase address.
+          const inputAddress =
+            this.network.type.chainID === 30
```
gamalielhere, line 339: import and reference RSK type instead of hardcoding.

gamalielhere (changes requested):

src/modules/address-book/ModuleAddressBook.vue
ahsan-javaiid: Done ✅

gamalielhere:
will be included in our release.

PR #5040, opened 2024-11-13
fix: lock to usd or browser setting

gamalielhere:
### Fix

* \[x] Add PR label


mfxrhxn (commented):
pybit/exceptions.py

PR #5110, opened 2025-02-27
chore: update @trezor/connect-web to version 9.5.0

martykan:
Hello from Trezor,

We noticed that you are using an older version of the @trezor/connect-web library, which we are planning to deprecate in the upcoming months.

This PR updates it to the latest version to ensure continued compatibility.


gamalielhere:
Moved to a different PR.

PR #5266, opened 2026-01-12
Fix transaction gas param before sendTransaction 

ahsan-javaiid:
### Fix

  * \[ ] Link to issue:
* \[ ] Add PR label

## Description:

When i try to swap RIF -> USDC and click send transaction then i am facing the error: 
```
Missing parameter, gasPrice, gas or value
```

The reason is that gas field in tx payload is coming as 0x0 instead of null and causing `estimateGas` to bypass in the following conditional clause: tx.gas = !tx.gas ? await ethCalls.estimateGas(localTx) : tx.gas;

This pull request fixes this issue by including "0x0" into the conditional clause so that the estimateGas call could trigger. 

### How to reproduce the issue

https://github.com/user-attachments/assets/f80afa01-0d19-451e-9a0a-fe019c4562d8

### DIFF
--- a/changelog/fix-5266.md
+++ b/changelog/fix-5266.md
@@ -1,0 +1,1 @@
+fix: fix tx gas parameter in send transaction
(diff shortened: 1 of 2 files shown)


olgakup:
Hi, we are not making anymore changes to this version of the app.

ahsan-javaiid:
@olgakup Would be great if you can share the correct version or branch so i can send the fix for the correct version. Thank you!

PR #5328, opened 2026-03-30
feat: ✨ sentry version tracking enabled

nicoirigoitia:
### Feature

Description:

Adds release field to Sentry initialization so errors are associated with the app version from package.json. This enables version/release tracking in the Sentry dashboard — filtering errors by release, detecting regressions, and identifying which version introduced a bug.

## Summary by CodeRabbit

* **Chores**
  * Added application version tracking and configured the app to include build version information with error reports and runtime configuration. This ensures the running build version is propagated from build metadata into the app, improving diagnostics and consistency for troubleshooting and monitoring.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -29,6 +29,7 @@ if (dsn) {
   sentryInit({
     app,
     dsn,
+    release: 'myetherwallet@' + configs.APP_VERSION,
     integrations: [
       browserTracingIntegration({ router }),
       replayIntegration(),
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -29,6 +29,7 @@ if (dsn) {
   sentryInit({
     app,
     dsn,
+    release: `myetherwallet@${configs.APP_VERSION}`,
```
olgakup: can you just keep it as  `myetherwallet@${configs.APP_VERSION}`

olgakup (changes requested):

src/main.ts
nicoirigoitia: done

gamalielhere (approved):

olgakup (approved):

PR #5333, opened 2026-03-31
feat: ✨ add filter low fiat balance tokens

gamalielhere:
## Summary by CodeRabbit

* **New Features**
  * Added a "Hide small balances" option in the balances table menu to filter out tokens with zero/low fiat value, keeping your portfolio view cleaner; can be toggled on/off.
* **Documentation**
  * Updated changelog headings and added an entry for the new low-balance token filter.


gamalielhere:
Already in feat/v7-develop

PR #5423, opened 2026-04-28
feat: paginate perps Markets, Positions and Fills tables

renerpdev:
## What

Adds pagination to the long perps tables so users can browse them comfortably instead of scrolling through every row at once.

Tables included:
- **Markets** (Perpetuals Markets list)
- **Positions** tab (My Positions)
- **Fills** tab (My Fills)

Tables intentionally **left as-is** (out of scope for this ticket):
- Orders
- Deposits & Withdrawals

## How it looks

Each paginated table now shows a small footer at the bottom-right with chevron arrows and a page indicator (`1 of 5` for client-side, `Page 1` for the cursor-based Fills table). Page size is **10 rows per page**.

## Behavior notes

- Markets and Positions paginate locally over the already-loaded list, so navigating pages is instant. Filtering or searching resets to page 1 automatically when the new result set is shorter.
- Fills uses the API's cursor-based pagination so we only load 10 rows at a time. Auto-refresh still runs every 10 seconds, but only refreshes page 1 — if you've navigated to page 3, your view stays put.

## Scope

All new shared pieces (the pagination component and the cursor-pagination helper) live inside `src/modules/perps/`. We did **not** touch the existing pagination in non-perps tables (e.g. Portfolio Gains/Losses) — that migration is a follow-up.

## Ticket

[MEW-1615](https://myetherwallet.atlassian.net/browse/MEW-1615)

## Test checklist

- [ ] Markets table: navigate pages, change filter (All / Stocks / Watchlist / etc.), verify page resets correctly
- [ ] Markets table: search for a token; pagination updates to match results
- [ ] Positions tab: open ≥11 positions in sandbox, verify pagination appears and works
- [ ] Fills tab: navigate forward/back through pages; verify auto-refresh doesn't yank you off page 2+
- [ ] Orders and Deposits & Withdrawals tabs: verify they look unchanged

## Summary by CodeRabbit

* **New Features**
  * Pagination added to markets, positions, and fills lists with prev/next controls and page status; defaults to 10 items per page.
  * New pagination control offers smooth scroll back to the table when changing pages.

* **Improvements**
  * Fills switched to cursor-based pagination; polling limited to the first page and refresh/reset behavior improved on auth changes. Navigation disables while loading.

### DIFF
--- a/src/modules/perps/composables/useCursorPaginate.ts
+++ b/src/modules/perps/composables/useCursorPaginate.ts
@@ -13,18 +13,23 @@ export function useCursorPaginate<T>(
   const currentPage = ref(0)
   const cursorStack = ref<(string | undefined)[]>([undefined])
   const nextCursor = ref<string | undefined>(undefined)
+  // Bumped on reset() so any in-flight request started under a previous
+  // auth/wallet context is dropped instead of overwriting the new state.
+  let requestEpoch = 0
 
   async function fetchPage(cursor: string | undefined): Promise<boolean> {
+    const epoch = requestEpoch
     loading.value = true
     try {
       const res = await fetcher({ limit: pageSize, cursor })
+      if (epoch !== requestEpoch) return false
       items.value = res.result ?? []
       nextCursor.value = res.pageInfo?.nextCursor
       return true
     } catch {
       return false
     } finally {
-      loading.value = false
+      if (epoch === requestEpoch) loading.value = false
     }
   }
 
@@ -58,10 +63,12 @@ export function useCursorPaginate<T>(
   }
 
   function reset() {
+    requestEpoch += 1
     items.value = []
     cursorStack.value = [undefined]
     currentPage.value = 0
     nextCursor.value = undefined
+    loading.value = false
   }
 
   const hasNext = computed(() => !!nextCursor.value)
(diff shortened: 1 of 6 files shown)


src/modules/perps/composables/useCursorPaginate.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  async function refetch() {
+    const ok = await fetchPage(undefined)
+    if (ok) {
+      cursorStack.value = [undefined]
+      currentPage.value = 0
+    }
+  }
+
+  async function nextPage() {
```
olgakup, line 44: can you add scroll to top of table when a user is going to the next page

olgakup (changes requested):

renerpdev:
### Note on Fills table pagination format

The Fills table renders **"Page X"** while the Markets and Positions tables render **"X of N"**. This is intentional and stems from a difference in pagination strategy:

- **Markets / Positions** use client-side array pagination (`usePaginate.ts`). The full dataset is in memory, so `totalPages = Math.ceil(length / pageSize)` can be computed → `"1 of N"`.
- **Fills** use server-side cursor pagination (`useCursorPaginate.ts`) against the Ondo API. The endpoint returns the current page plus a `nextCursor` but **does not return a total record count**. Without a total we cannot compute `totalPages`, so `PerpsPagination.vue` falls back to its `"Page {{ currentPage + 1 }}"` branch.

If visual consistency is required, options are:
1. Hide the page indicator on cursor-paginated tables (just chevrons).
2. Ask the API team to include a total count in the fills response so we can show `"X of N"` everywhere.

Leaving as-is for this PR since it reflects a real API constraint, not a UI bug. Happy to file a follow-up ticket if PM wants it unified.

olgakup (approved):

PR #5439, opened 2026-05-04
feat: [MEW-1670] block letters in withdraw amount + reuse perps amount input

renerpdev:
## What was wrong

In the **Withdraw USDC** dialog, the *Amount* field was a plain text input. You could type letters into it (e.g., `abc`, `1.5x`), and there was no limit on how many decimal places you could enter. The TP/SL dialog already had a nice numeric-only input that blocks letters at the keystroke — that component was just hardcoded to a "Price" label, so it couldn't be reused elsewhere.

## What's new

- The TP/SL price input has been generalized into a reusable **PerpsAmount** component with a configurable title (default still `"Price"`, so TP/SL renders identically).
- The **Withdraw USDC** dialog now uses this component for the *Amount* field, so:
  - Letters and other non-numeric characters are blocked at the keystroke — they can no longer be typed.
  - Only one decimal point is allowed.
  - Entering more than **6** decimal places (USDC's max precision on mainnet) shows an inline error.
  - Entering an amount above the available balance shows an inline error and keeps the Withdraw button disabled.
- The **MAX** button now lives directly under the amount input, styled to match the percentage pills used elsewhere in the perps module.

## How to test

1. Open the perps Portfolio → click **Withdraw**.
2. In the *Amount* field:
   - Try typing letters (`abc`, `xyz`) → nothing should appear.
   - Try typing two decimal points (`1..5`) → the second `.` should be blocked.
   - Type `1.1234567` (7 decimals) → an error message appears: "Amount supports up to 6 decimal places".
   - Type a value larger than your available balance → "Amount exceeds available balance" appears, Withdraw button stays disabled.
   - Click **MAX** → field fills with your full available balance.
3. Open the **Auto Close (TP/SL)** dialog on a position and confirm the price inputs render and behave exactly as before — the percentage pills (+10%, +20%, etc. and -1% through -5%) should still set the price.

## Jira

https://myetherwallet.atlassian.net/browse/MEW-1670

## Summary by CodeRabbit

* **New Features**
  * Configurable input labels for perps amount fields
  * Unified amount input component used across Take Profit, Stop Loss and Withdraw dialogs
  * MAX control moved into the amount input footer for quick selection

* **Bug Fixes**
  * Stronger amount validation with precise decimal checks and clearer error messages
  * Resetting and consistent handling of amount state when dialogs open/submit

### DIFF
--- a/src/modules/perps/components/PerpsWithdrawDialog.vue
+++ b/src/modules/perps/components/PerpsWithdrawDialog.vue
@@ -27,29 +27,32 @@
 
         <!-- Amount -->
         <div class="mb-4">
-          <label
-            class="text-info text-s-11 uppercase tracking-wider font-medium mb-1.5 block"
+          <perps-amount
+            v-model:amount="amount"
+            v-model:error="amountError"
+            title="Amount"
+            :validate-input="validateAmount"
           >
-            Amount
-          </label>
-          <div class="flex items-center gap-2">
-            <input
-              v-model="amount"
-              type="text"
-              placeholder="0.00"
-              class="flex-1 bg-surface rounded-12 px-4 py-3 text-s-14 outline-none focus:ring-2 focus:ring-primary"
-            />
-            <button
-              class="text-primary text-s-13 font-medium"
-              @click="amount = withdrawableMargin"
-            >
-              MAX
-            </button>
-          </div>
+            <template #footer>
+              <div class="flex justify-start mt-1">
+                <button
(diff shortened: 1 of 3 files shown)


src/modules/perps/components/PerpsWithdrawDialog.vue
```diff
@@ -27,29 +27,32 @@
...
+                  MAX
+                </button>
+              </div>
+            </template>
+          </perps-amount>
         </div>
 
         <!-- Error -->
-        <p v-if="error" class="text-error text-s-12 mb-4">{{ error }}</p>
+        <p
```
olgakup: oh I think i forgot to add this in actual Amount Input can you moved it to the amount for model value "amountError || error"

and can you add transition fade to it

olgakup (changes requested):

olgakup (approved):

PR #5486, opened 2026-05-15
feat: [MEW-1805] add tooltip to time in force order detail

renerpdev:
## What's new

The perps order details dialog now shows a small info-icon next to the **Time in Force** label. Hovering it reveals a tooltip that explains what the field means:

> The order remains active until the order is completely filled or you manually cancel it.

This addresses the QA question — users could see the "Time in Force" value but had no in-app explanation of the underlying policy.

## How to test

1. Open MEW perps and place (or look up) any order that has a `Time in Force` value — currently the only order type where this field is shown.
2. Open the order detail dialog from the orders list.
3. Hover the small info icon next to "Time in Force".
4. Confirm the tooltip text reads exactly: *The order remains active until the order is completely filled or you manually cancel it*.
5. Other rows (Market, Side, Status, etc.) should not show the info icon.

## New Features
* Order dialog fields now display contextual tooltips for additional guidance
* "Time in Force" field includes a tooltip explaining when orders remain active

[Review Change Stack](https://app.coderabbit.ai/change-stack/MyEtherWallet/MyEtherWallet/pull/5486?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/modules/perps/components/PerpsOrderDialog.vue
+++ b/src/modules/perps/components/PerpsOrderDialog.vue
@@ -37,10 +37,17 @@
             :key="row.label"
             class="flex items-center justify-between px-5 py-4"
           >
-            <span
-              class="text-s-11 uppercase text-info tracking-sp-06 font-bold"
-              >{{ row.label }}</span
-            >
+            <div class="flex items-center gap-1">
+              <span
+                class="text-s-11 uppercase text-info tracking-sp-06 font-bold"
+                >{{ row.label }}</span
+              >
+              <app-tooltip
+                v-if="row.tooltip"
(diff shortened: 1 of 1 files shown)


node_modules
```diff
@@ -0,0 +1 @@
+../../MyEtherWallet/node_modules
```
olgakup: whats this change? this folder should not be here, it should be in gitignore

olgakup (changes requested):

node_modules
renerpdev: sorry this is a mistake, will fix asap

olgakup (approved):

PR #5509, opened 2026-06-02
[MEW-1821] gate perps fetches behind isPerpsActive

renerpdev:
## Summary
- Adds `usePerpsActive` composable: a reactive flag = `route.startsWith('/perps') || (walletPanel === 'perps' && isOpenSideMenu)`
- Adds `gatedPoll(fn, ms)` helper that wraps `setInterval` to skip ticks when inactive and refire on every false→true transition
- Wires the gate into the five leaking module-level singletons: `usePerpsContracts`, `usePerpsMarkPrices`, `usePerpsPositions`, `usePerpsBalance`, `usePerpsPortfolioSummary`

## What was leaking
| Composable | Interval |
|---|---|
| `usePerpsContracts` | 2.5s |
| `usePerpsMarkPrices` | 5s |
| `usePerpsPositions` | 5s |
| `usePerpsBalance` | 1s |
| `usePerpsPortfolioSummary` | 5min |

All five were module-level singletons that never stopped. Now they no-op when `isPerpsActive === false`.

## What was NOT touched
- `usePerpsAuth` token-restoration `watch(walletAddress)` — runs only on address change, not periodic
- `usePerpsHistory` (`usePerpsOrders`, `usePerpsFills`, `usePerpsDepositsWithdrawals`) — already use `onUnmounted` cleanup
- `usePerpsPortfolioGraph` cache-clear interval — clears cache, no network
- `refreshKey` watchers in `usePerpsPositions` / `usePerpsBalance` — only fire after user mutations, which are inherently inside perps

## Test plan
- [x] `pnpm test:unit run tests/unit/specs/modules/perps/composables/usePerpsActive.spec.ts` — 10/10 pass
- [x] `pnpm test:unit run` (full suite) — 64/64 pass
- [x] `pnpm type-check` — clean
- [x] `pnpm lint` — clean
- [x] Manual: opening then closing the perps side panel stops fetching; navigating to `/stocks` shows 0 perps requests; navigating into `/perps` triggers an immediate refetch and resumes intervals

## Summary by CodeRabbit

* **New Features**
  * Introduced gated polling for the perps module so balances, positions, mark prices, markets and portfolio summaries poll only when the perps view/panel is active; polling now starts/stops with UI context and triggers immediately on activation.

* **Tests**
  * Added unit tests covering perps-active detection and gated polling behavior across navigation and menu state changes.

### DIFF
--- a/src/modules/perps/composables/usePerpsAuth.ts
+++ b/src/modules/perps/composables/usePerpsAuth.ts
@@ -4,6 +4,7 @@ import { useWalletStore } from '@/stores/walletStore'
 import { storeToRefs } from 'pinia'
 import type { PerpsBalance, PortfolioSummary } from '../sdk/types'
 import { usePerpsToasts } from '@/modules/perps/composables/usePerpsToasts'
+import { gatedPoll } from '@/modules/perps/composables/usePerpsActive'
 import { decrypt, encrypt } from '@/utils/crypto'
 
 const STORAGE_KEY_TOKEN = 'perps_auth_token'
@@ -216,7 +217,7 @@ export function usePerpsBalance() {
     }
 
     poll()
-    setInterval(poll, 1_000)
+    gatedPoll(poll, 1_000)
 
     let lastRefreshKey = refreshKey.value
     setInterval(() => {
@@ -286,7 +287,7 @@ export function usePerpsPortfolioSummary() {
     }
 
     poll()
-    setInterval(poll, 300_000) // refresh every 5 minutes
+    gatedPoll(poll, 300_000) // refresh every 5 minutes
 
     _sharedFetchSummary = fetchSummary
   }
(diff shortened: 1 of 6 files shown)


src/modules/perps/composables/usePerpsAuth.ts
```diff
@@ -216,7 +217,7 @@ export function usePerpsBalance() {
     }
 
     poll()
-    setInterval(poll, 1_000)
+    gatedPoll(poll, 1_000)
```
olgakup, line 220: can you make it every 3 seconds

src/modules/perps/composables/usePerpsMarkets.ts
```diff
@@ -70,7 +71,7 @@ export function usePerpsContracts() {
   if (!contractsInitialized) {
     contractsInitialized = true
     fetchContracts()
-    setInterval(fetchContracts, 2500)
+    gatedPoll(fetchContracts, 2500)
```
olgakup, line 74: 3 seconds please

olgakup (changes requested):

gamalielhere (approved):

renerpdev:
since im working on the ws adoption i will include the gating there, so we dont have stale work of this PR

PR #5512, opened 2026-06-03
Feat/v7 o perps live

gamalielhere:
## Summary by CodeRabbit

* **Style**
  * Minor formatting improvement to the login payload.

* **Chores**
  * Updated perps endpoint configuration so websocket/HTTP connectivity now targets the live environment when not in dev mode, while using the sandbox endpoint during development.


olgakup:
can you add a button somewhere to switch it  when testing, so its controlled what we are testing. We will remove it later

PR #5514, opened 2026-06-04
fix: 🐛 trap keyboard focus inside open dialogs

nicoirigoitia:
## What
Add focus trapping to dialogs so Tab key navigation stays within the open modal instead of cycling through background elements.

## Why
When a dialog was open, pressing Tab allowed navigation through buttons behind the overlay, breaking keyboard accessibility expectations.

## How
- Add `:inert` attribute to the app's main content wrapper when a dialog is open, preventing Tab from reaching background elements
- Fix the dialog watch to always sync `isAreaHidden` with `isOpen` state (removed a buggy `focused.value` condition that caused timing issues)
- Auto-focus the dialog container on open using `nextTick` to ensure the DOM is ready
- Remove unused `useFocus` import that was no longer needed

## Summary by CodeRabbit

* **Bug Fixes**
  * App main area now becomes inert/disabled when overlay panels are hidden to prevent background interaction.
  * Dialog open/close flow improved: when a dialog opens it waits for DOM updates and moves focus into the dialog, and it reliably updates background interaction state for better accessibility.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="relative h-screen overflow-hidden">
+  <div class="relative h-screen overflow-hidden" :inert="isAreaHidden || undefined">
     <welcome-dialog v-if="!isDevMode" />
     <the-app-layout v-if="isLoadingComplete" :aria-hidden="isAreaHidden" />
     <module-toast />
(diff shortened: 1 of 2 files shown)


renerpdev (approved):
LGTM

gamalielhere (approved):

PR #5515, opened 2026-06-04
fix: MEW-1856 restrict perps network dropdown to ETH only

renerpdev:
## Summary
- When the user is on a perps route, the topbar `TheCurrentNetwork` chain dialog now lists only Ethereum (other routes are unaffected).
- The perps page (`ViewPerps`) and perps drawer (`ModulePerpsTrade`) "Switch to Ethereum" dialogs also only show Ethereum, removing the dead-end of selecting another non-ETH chain when the user is explicitly being told that perps requires ETH.

All three dropdowns reuse the existing `passedChains` filter that `SelectChainDialog` already supports — no new prop or composable was added.

## Test plan
- [x] `pnpm vue-tsc --noEmit` clean
- [x] `pnpm test:unit --run` — 56 / 56 pass
- [x] Topbar dropdown on `/perps` shows only Ethereum
- [x] Topbar dropdown on `/perps/perp/:market` shows only Ethereum
- [x] Topbar dropdown on `/` / `/crypto` still shows the full chain list (regression)
- [x] Perps page "Switch to Ethereum" dialog shows only Ethereum
- [x] Perps drawer "Switch to Ethereum" dialog shows only Ethereum

## Summary by CodeRabbit

* **Bug Fixes**
  * Perps trading network selection now restricts available chains to Ethereum on Perps pages, preventing attempts on unsupported networks.

* **Chores**
  * Centralized Ethereum-only chain handling applied across Perps views and the app layout so the network selector behaves consistently.

### DIFF
--- a/src/components/core_layouts/wallet/TheCurrentNetwork.vue
+++ b/src/components/core_layouts/wallet/TheCurrentNetwork.vue
@@ -1,6 +1,6 @@
 <template>
   <!-- App Select Chain -->
-  <select-chain-for-app :has-label="false">
+  <select-chain-for-app :has-label="false" :passed-chains="restrictedChains">
     <template #network-button="{ openNetworkDialog, selectedChain }">
       <button
         class="hoverNoBG p-1 xs:py-2 xs:px-3 rounded-[24px] xs:rounded-full max-w-[178px] shadow-button shadow-button-elevated"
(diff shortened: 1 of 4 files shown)


src/components/core_layouts/wallet/TheCurrentNetwork.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <!-- App Select Chain -->
-  <select-chain-for-app :has-label="false">
+  <select-chain-for-app :has-label="false" :passed-chains="restrictedChains">
```
olgakup, line 3: I would not change this based on the route

olgakup:
Not applicable

PR #5546, opened 2026-06-15
fix: [MEW-1927] guard swap gas-fee quote against empty transactions

renerpdev:
## What was wrong

After choosing a swap quote, the app was occasionally throwing an unhandled "Invalid request body." error from the gas-fee quote step. This generated ~1735 reports in Sentry over the last ~2 months and was visible to the user as a silent failure (and a Sentry alert for us).

## What changed

In `src/modules/swap/ModuleSwap.vue`:

- `generateEVMGasFeeQuote` and `generateBTCGasFeeQuote` now early-return when there are no transactions to quote, instead of posting an empty list to the backend (which the backend rightly rejects with a 400).
- The watcher that fires after a quote is selected is now wrapped in try/catch, so any unexpected error during quote fetching is reported to Sentry but does not bubble up as an unhandled rejection.

User experience: the existing toast that fires when `getSwap` fails still triggers; there is no new UI.

## How to test

1. Open the app, connect a wallet, go to the Swap module.
2. Pick a from-token and to-token on Ethereum, enter an amount that produces quotes.
3. Pick a provider from the quote list and let the gas-fee step run.
4. Repeat across a couple of token pairs (including one where you previously saw the error, if you have one). Mobile Safari on iOS is the original-reported environment.

Expected: no "Invalid request body." showing up in the console / Sentry; swap flow proceeds normally. If the backend does reject the gas-fee request for any reason, the existing "something went wrong" toast appears instead of an unhandled rejection.

## Sentry

https://myetherwallet-inc.sentry.io/issues/7431687868/

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined gas fee quote generation to prevent unnecessary calculations when no underlying transactions are detected, improving efficiency.
  * Enhanced swap operation error handling with improved exception tracking, state management, and recovery mechanisms during the process.

### DIFF
--- a/src/modules/swap/ModuleSwap.vue
+++ b/src/modules/swap/ModuleSwap.vue
@@ -1460,22 +1460,32 @@ watch(
 
 watch(
   () => selectedQuote.value,
-  async provider => {
+  async (provider, _prev, onCleanup) => {
     if (!provider) return
+    // A newer selectedQuote invalidates this run: flag it so late-resolving
+    // getSwap / gas-fee-quote calls don't overwrite state with stale data.
+    let cancelled = false
+    onCleanup(() => {
+      cancelled = true
+    })
     // disable proceeding while we fetch swap info for the selected quote to prevent user from clicking "proceed" before we have the necessary transaction data
     txProceeding.value = true
     try {
       const res = await getSwap(provider)
+      if (cancelled) return
       swapInfo.value = res
       const quoteRes = await (isBitcoinChain.value
         ? generateBTCGasFeeQuote()
         : generateEVMGasFeeQuote())
+      if (cancelled) return
       swapGasFeeQuote.value = (quoteRes as QuotesResponse) || undefined
     } catch (err) {
+      if (cancelled) return
+      swapInfo.value = null
       swapGasFeeQuote.value = undefined
       captureException(err)
     } finally {
-      txProceeding.value = false
+      if (!cancelled) txProceeding.value = false
     }
   },
   { deep: true },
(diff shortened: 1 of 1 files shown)


src/modules/swap/ModuleSwap.vue
```diff
@@ -1461,16 +1461,20 @@ watch(
...
-      // disable proceeding while we fetch swap info for the selected quote to prevent user from clicking "proceed" before we have the necessary transaction data
-      txProceeding.value = true
-      await getSwap(provider).then(async res => {
-        swapInfo.value = res
-        const quoteRes = await (isBitcoinChain.value
-          ? generateBTCGasFeeQuote()
-          : generateEVMGasFeeQuote())
-        swapGasFeeQuote.value = (quoteRes as QuotesResponse) || undefined
-      })
+    if (!provider) return
```
gamalielhere, line 1464: This may be breaking the feature. Please double check and confirm.

renerpdev: not sure if i understand, this is just an early exit, the opposite of the conditional that we had before. could explain what do you mean by breaking the feature? ty

gamalielhere (approved):

src/modules/swap/ModuleSwap.vue
```diff
@@ -1456,16 +1461,20 @@ watch(
...
+      const res = await getSwap(provider)
+      swapInfo.value = res
+      const quoteRes = await (isBitcoinChain.value
+        ? generateBTCGasFeeQuote()
+        : generateEVMGasFeeQuote())
+      swapGasFeeQuote.value = (quoteRes as QuotesResponse) || undefined
+    } catch (err) {
+      swapGasFeeQuote.value = undefined
+      captureException(err)
+    } finally {
```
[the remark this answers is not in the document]
renerpdev: Done in d616f3eccf — the `selectedQuote` watcher now takes `onCleanup`, sets a `cancelled` flag, and skips the `swapInfo` / `swapGasFeeQuote` / `txProceeding` writes when a newer quote has superseded the run, so a late-resolving `getSwap`/gas-fee-quote can't overwrite state.

PR #5549, opened 2026-06-17
fix: [MEW-1937] handle European locale comma and invalid values in parseUnits

renerpdev:
## What was wrong

Users with European locales (Poland, Germany, France, etc.) type commas as decimal separators — e.g. `0,125` instead of `0.125`. viem's `parseUnits` only accepts periods, so it throws `InvalidDecimalNumberError` and crashes the Send module. A secondary variant occurs when a token's `balance` field contains a non-decimal string (hex hash from a malformed API response).

566 occurrences since April 2026 (Sentry APP-MEW-WEB-W5).

## What changed

- **AppEnterAmount.vue**: Comma keypress is now auto-converted to a period instead of being blocked. Pasted text with commas is also normalized to periods.
- **ModuleSend.vue**: All `parseUnits` calls are wrapped in a `safeParseUnits` helper that normalizes commas→periods and validates the input is a valid decimal before parsing. Invalid values fall back to `BigInt(0)` instead of throwing.

## How to test

1. Open the Send module (connect any wallet on Ethereum)
2. Type an amount using a comma as decimal separator (e.g. `0,125`) — it should auto-convert to `0.125`
3. Paste a value with commas (e.g. `1,5`) — should normalize to `1.5`
4. Verify normal decimal input (`0.5`, `100`) still works as expected
5. Verify balance validation (insufficient balance) still triggers correctly

Sentry: https://myetherwallet-inc.sentry.io/issues/7432800886/

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved amount entry handling by converting comma decimal separators to periods during typing and paste, while preserving cursor/selection position.
  * Strengthened send amount parsing and validation to safely handle empty or malformed inputs, improving correctness for both token transfers and BTC sends.

### DIFF
--- a/src/components/AppEnterAmount.vue
+++ b/src/components/AppEnterAmount.vue
@@ -81,6 +81,7 @@ import {
   formatFiatValue,
 } from '@/utils/numberFormatHelper'
 import { useInFocusInput } from '@/composables/useInFocusInput'
+import { useNumericInput } from '@/composables/useNumericInput'
 
 const walletStore = useWalletStore()
 const { isLoadingBalances: isLoading, isWalletConnected } =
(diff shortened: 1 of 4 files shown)


nicoirigoitia:
LGTM, clean refactor. Just fix the regex issue CodeRabbit flagged before merging (".", "-", "-." still pass validation) and we're good

nicoirigoitia (approved):

gamalielhere (approved):

PR #5560, opened 2026-06-23
fix: 🐛 sentry error forbidden

nicoirigoitia:
## What
Stop reporting expected 4xx client errors (e.g. the 403 from purchase/info) to Sentry from the centralized useMEWFetch wrapper.

## Why
Region-restricted endpoints like v5/purchase/info return 403, which the fetch error handler was sending to Sentry as "Error: Forbidden", generating ~663 noise events with no actionable signal.

## How
- Detect client errors (status 400–499) in useMEWFetch's onFetchError handler
- Skip captureException for those 4xx responses
- Keep reporting network and 5xx errors that exhaust retries
- No functional change: ctx.error is still set, so callers' catch blocks behave as before

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fetch error handling by differentiating client-side (4xx) vs other failures, reducing unnecessary monitoring noise.
* **New Features**
  * Added support for showing and sorting using trending token addresses during swaps.
* **UI Improvements**
  * Enhanced the token selection dialog with improved search/sort layout, better focus behavior, and richer sorting based on market metrics (when available).


gamalielhere:
moved to #5601

PR #5630, opened 2026-07-16
[MEW-2047] fix: localize AppSearchInput placeholder default

renerpdev:
## Summary

The `AppSearchInput` `placeholder` prop defaulted to the hardcoded English literal `'Search'`, so any caller that omits `:placeholder` rendered an untranslated `Search` regardless of the active locale. This was visible in the **Connect Wallet** dialog's wallet search (reported) and also in the **Stocks list** search — both omit the prop.

## Fix

- `src/components/AppSearchInput.vue`: default the `placeholder` prop to `''` and fall back to `$t('common.search')` for both `:placeholder` and `:aria-label`. The `common.search` key already exists (`= "Search"`), so English is unchanged and all other locales now localize correctly.
- Callers that pass an explicit `:placeholder` are unaffected (their value overrides the fallback).

## Tests

- Adds `tests/unit/components/AppSearchInput.spec.ts` (the repo's first component-mount spec): localized fallback in English, localized fallback in a non-English locale (regression guard against the hardcoded default), explicit-override, and aria-label mirroring. **4/4 passing.**
- `vue-tsc --noEmit` → 0 errors · `eslint` → 0 errors.

## Out of scope

`TableTokenBalance.vue` passes a hardcoded literal `placeholder="table options"` (supplies an explicit value, so unaffected by this change). Logged as a follow-up.

## Testing

1. Switch the app language to a non-English locale (Settings → language dropdown).
2. Open **Connect Wallet** → the wallet-grid search placeholder is translated (no longer `Search`).
3. In English the placeholder still reads `Search` (no regression).

## Summary by CodeRabbit

* **Accessibility**
  * Search inputs now use localized labels and placeholders, including support for language changes.

* **Bug Fixes**
  * Fixed the default search text so it no longer remains hardcoded in English.
  * Explicit placeholder text continues to override the localized default.

### DIFF
--- a/src/components/AppSearchInput.vue
+++ b/src/components/AppSearchInput.vue
@@ -18,8 +18,8 @@
         size === 'compact' ? 'pl-10 text-[15px]' : 'pl-[46px] text-[17px]',
         bgClass,
       ]"
-      :aria-label="placeholder"
-      :placeholder="placeholder"
+      :aria-label="placeholder || $t('common.search')"
+      :placeholder="placeholder || $t('common.search')"
       @focus="inFocusInput = true"
       @blur="inFocusInput = false"
     />
@@ -60,10 +60,11 @@ import { XCircleIcon } from '@heroicons/vue/24/outline'
 defineProps({
   /**
    * @placeholder The placeholder text of the input field. Also used as the aria label.
+   * When empty, falls back to the localized `common.search` string.
    */
   placeholder: {
     type: String,
-    default: 'Search',
+    default: '',
   },
   /**
    * @bgClass The background color of the input field.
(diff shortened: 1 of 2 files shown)


renerpdev:
This PR also covers **[MEW-2048](https://myetherwallet.atlassian.net/browse/MEW-2048) — Search (Your Balance)**.

The reported untranslated `Search` placeholder on the **Your Balance** screen comes from the same root cause: the token-balance search input (`TableTokenBalance.vue`) renders `<app-search-input v-model="searchInput" />` **without** an explicit `:placeholder`, so it used the hardcoded English default. With this change it falls back to `$t('common.search')` and now localizes correctly (e.g. `Buscar` in Spanish, `搜索` in Chinese). No additional code change required — MEW-2048 is resolved by this same fix.

olgakup (approved):

gamalielhere (approved):

PR #5631, opened 2026-07-16
[MEW-2034] fix: localize Earn page email subscription error messages

renerpdev:
## Summary

The inline validation error shown **below the email input on the Earn page** (`/earn` → `ViewTemp.vue` → `AppSubscribeToUpdates`) was hardcoded in English and did not respond to the language selector.

## Changes

- Route the two inline validation messages in `useEmailSubscription.ts` through `vue-i18n` — `common.subscribe.email_required` and `common.subscribe.email_invalid`.
- Expose `emailErrorMessage` as a `computed` over an error-key ref so it **re-translates reactively** when the locale changes. Resolving `t()` into a plain `ref` froze the message in whatever locale was active at validation time (this was why the error stayed English after switching language).
- Add the new keys to every bundled locale: `en`, `es`, `zh`. Previously only `en.json` had them, so other locales fell back to the English source.

## Out of scope

- The three subscribe **toast** strings (`Thank you for subscribing!`, `Something went wrong`, `Please try again later.`) are intentionally left in English and will be localized in a separate follow-up PR.

## Testing

- New spec `tests/unit/composables/useEmailSubscription.spec.ts` (5 tests): key resolution in `en`/`es`, error clearing on a valid email, and **locale-switch reactivity** of an already-visible error.
- `pnpm vitest run` green · `pnpm vue-tsc --noEmit -p tsconfig.app.json` clean · `eslint` clean.
- Manual smoke on `/earn`: switching language now translates the inline error for both the empty and the invalid-email cases.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added localized email placeholder and validation messages in English, Spanish, and Chinese.
  * Email validation errors now update automatically when the application language changes.

* **Bug Fixes**
  * Improved handling of required and invalid email messages to ensure they display in the active locale.

* **Tests**
  * Added coverage for email validation, error clearing, and locale switching behavior.

### DIFF
--- a/src/composables/useEmailSubscription.ts
+++ b/src/composables/useEmailSubscription.ts
@@ -1,4 +1,4 @@
-import { ref } from 'vue'
+import { computed, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { watchDebounced } from '@vueuse/core'
 import { useToastStore } from '@/stores/toastStore'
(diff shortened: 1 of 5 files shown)


renerpdev:
### Heads-up: #5632 (MEW-2036) builds on this PR

#5632 (localize **all** toast messages) touches the same `src/composables/useEmailSubscription.ts` and `src/i18n/locales/common/{en,es,zh}.json`. Suggested order: **merge this PR first, then #5632 rebases on top.**

Note for whoever merges second: git auto-merges `useEmailSubscription.ts` **without a conflict marker** but leaves a duplicate `const { t } = useI18n()` + duplicate `vue-i18n` import that must be removed by hand (build break otherwise). The three `common/*.json` `subscribe`-block conflicts are the normal flagged kind — keep all keys from both sides. Full resolution recipe is on #5632.

gamalielhere (approved):

olgakup (approved):

PR #5681, opened 2026-07-29
fix: [MEW-2083] drop viem tx receipt timeout sentry noise

renerpdev:
## What was wrong

Sentry was collecting `WaitForTransactionReceiptTimeoutError` (viem) whenever a submitted trade/swap transaction did not confirm within viem's `waitForTransactionReceipt` timeout window. 153 events since April, 0 users impacted, mostly on mobile in-app browsers (Chrome Mobile WebView / MEW Mobile). This is an inherent blockchain/network condition — a slow node, congestion, a dropped tx, or a flaky mobile connection — not a defect in MEW code. The error is already `handled: yes` (the app catches it and the tx may still confirm later), so it was pure Sentry noise.

## What changed

Added a narrow `beforeSend` filter, `isTransactionReceiptTimeoutError`, in `src/sentry/extensionNoise.ts`, wired into `src/main.ts` alongside the existing extension / invalid-address / transient-RPC noise filters. It matches viem's guaranteed error `name` (`WaitForTransactionReceiptTimeoutError`), walking the `cause` chain so it survives minification and nesting. Kept deliberately separate from `isTransientRpcError` so trade quote-retry behavior is untouched. Added a unit spec.

Only Sentry reporting is affected — genuine app errors and all user-visible behavior are unchanged.

## How to test

This change is not directly user-visible (it only suppresses a Sentry report). To verify behavior is intact:
- Run the wallet, connect a wallet, submit a swap/trade on Ethereum, and confirm the flow behaves exactly as before (success toast on confirmation, existing handling on timeout).
- Unit: `pnpm vitest run tests/unit/sentry/extensionNoise.spec.ts` (15 tests pass, including the new `isTransactionReceiptTimeoutError` cases).
- Type-check + lint are green.

## Assumptions (full-auto sweep)

- Treated this as unactionable Sentry noise rather than a behavioral bug: the error is `handled: yes` and the timeout is intrinsic to on-chain confirmation, so there is no MEW-side code path that can prevent it. If the team instead wants a user-facing "still pending" affordance on receipt timeout, that is a separate feature ticket.

## Sentry

https://myetherwallet-inc.sentry.io/issues/APP-MEW-WEB-X6

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary error reporting by filtering transaction receipt timeout errors from diagnostics.
  * Improved handling of application state captured in diagnostic reports to avoid sending oversized chain data.
* **Tests**
  * Added coverage for timeout errors represented as native errors, serialized payloads, nested causes, and unrelated inputs.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -20,6 +20,7 @@ import configs from '@/configs'
 import {
   isExtensionOrProviderError,
   isInvalidWalletAddressError,
+  isTransactionReceiptTimeoutError,
 } from '@/sentry/extensionNoise'
 import { isTransientRpcError } from '@/modules/trade/composables/transientRpcError'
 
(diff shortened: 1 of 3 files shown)


gamalielhere (approved):

renerpdev:
Consolidated into #5683 — all four Sentry noise-filter PRs (#5647, #5681, #5682, #5683) touched the same two files and conflicted with each other, so `isTransactionReceiptTimeoutError` (MEW-2083) now lives in #5683 alongside the other predicates, rebased on latest `feat/v7-develop`. Closing this in favor of #5683.

PR #5699, opened 2026-08-06
fix: keep header More dropdown above the trade drawer (MEW-2113)

renerpdev:
## Summary

The header "More" (Más) dropdown was rendered **behind** the right-side trade drawer ("Operar"), clipping its items.

### Root cause
`TheHeader.vue` is `fixed z-10`, which makes it its own stacking context. The dropdown popover lives inside `AppSelect.vue` with `absolute z-[500]`, but that `500` only ranks it *within the header's* `z-10` context. The trade drawer (`LayoutWallet.vue`) is a **sibling** of the header with `z-[49]/z-[50]/z-[51]` — all `> 10` — so the whole drawer paints in front of the whole header, covering the dropdown regardless of its internal `z-[500]`.

### Fix
Reuse the pattern the header already uses for global search (`isSearchOpen ? 'z-[201]' : 'z-10'`): lift the whole header above the drawer while a header dropdown is open.

- `AppSelect.vue`: `openSelect` is now exposed as `v-model:open` (`defineModel`). Additive and defaulted to `false`, so existing callers are unaffected.
- `TheHeader.vue`: binds `v-model:open` on the More and Learn dropdowns and lifts the header to `z-[201]` via `isHeaderOverlayOpen` (search **or** More **or** Learn open).

Learn is covered too since it is the same component in the same stacking trap.

## Testing
- `tests/unit/components/AppSelect.spec.ts` (new, 2/2) — guards the `v-model:open` contract the header relies on.
- `vue-tsc --noEmit -p tsconfig.app.json` → exit 0.
- eslint on changed files → clean.
- Manual: open the "Operar" drawer, then hover/click "More" — dropdown renders fully above the drawer, all items visible/clickable, including below the responsive thresholds where Learn/Earn fold into "More". Global search overlay still works.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the “More” header dropdown appearing behind the trade drawer.
  * Improved header layering when Search, Learn, or More menus are open.

* **Enhancements**
  * Added support for observing and controlling dropdown open states.
  * Ensured dropdowns close correctly after selecting an option.

* **Tests**
  * Added coverage for dropdown opening, closing, and state updates.

### DIFF
--- a/changelog/fix-5699.md
+++ b/changelog/fix-5699.md
@@ -1,0 +1,1 @@
+- Fixed the header "More" dropdown being hidden behind the trade drawer (MEW-2113)
--- a/changelog/fix-5699.md
+++ b/changelog/fix-5699.md
@@ -1,1 +1,0 @@
-- Fixed the header "More" dropdown being hidden behind the trade drawer (MEW-2113)
--- a/src/components/AppSelect.vue
+++ b/src/components/AppSelect.vue
@@ -189,9 +189,12 @@ const targetValue = ref<HTMLElement | null>(null)
  */
 const selected = defineModel<AppSelectOption>('selected', { required: false })
 /**
- * controls the open state of the select dropdown
+ * Open state of the dropdown, exposed as `v-model:open` so an ancestor can
+ * react to it. TheHeader uses this to lift its z-index while a header dropdown
+ * is open, otherwise the popover is clipped behind sibling drawers (MEW-2113).
+ * Defaults to false, so callers that don't bind `open` are unaffected.
  */
-const openSelect = ref(false)
+const openSelect = defineModel<boolean>('open', { default: false })
 
 /**
  * @method toggleSelect
--- a/src/components/core_layouts/TheHeader.vue
+++ b/src/components/core_layouts/TheHeader.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     class="flex items-center w-full h-[68px] sm:h-[76px] fixed top-0 px-5 md-header:px-5 bg-white shadow-[0px_3px_12px_-6px_rgba(0,0,0,0.32)]"
-    :class="isSearchOpen ? 'z-[201]' : 'z-10'"
+    :class="isHeaderOverlayOpen ? 'z-[201]' : 'z-10'"
   >
     <div class="flex w-full justify-between items-center mx-auto gap-3">
       <!-- LOGO -->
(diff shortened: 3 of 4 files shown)


changelog/fix-5699.md
```diff
@@ -0,0 +1 @@
+- Fixed the header "More" dropdown being hidden behind the trade drawer (MEW-2113)
```
gamalielhere: no need for changelog

gamalielhere (approved):

changelog/fix-5699.md
renerpdev: Good catch — removed the changelog entry in d35bddf0b5. Thanks!

src/components/AppSelect.vue
```diff
@@ -189,9 +189,12 @@ const targetValue = ref<HTMLElement | null>(null)
...
 const selected = defineModel<AppSelectOption>('selected', { required: false })
 /**
- * controls the open state of the select dropdown
+ * Open state of the dropdown, exposed as `v-model:open` so an ancestor can
+ * react to it. TheHeader uses this to lift its z-index while a header dropdown
+ * is open, otherwise the popover is clipped behind sibling drawers (MEW-2113).
+ * Defaults to false, so callers that don't bind `open` are unaffected.
  */
-const openSelect = ref(false)
+const openSelect = defineModel<boolean>('open', { default: false })
```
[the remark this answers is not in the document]
renerpdev: Skipping this one — the scenario doesn't occur with how the component is actually used.

`targetValue` is set on every path that opens the dropdown: `toggleSelect()` (button click) and the hover `watch` both assign `targetValue.value = target.value`. Those are the only ways a dropdown opens here.

In `TheHeader.vue` the `v-model:open` bindings (`isMoreOpen`/`isLearnOpen`) are **observe-only** — they're `ref(false)` read solely by the `isHeaderOverlayOpen` computed to lift the z-index, and are never written to `true` from the parent. The open always originates inside `AppSelect`, so `targetValue` is always populated and `onClickOutside` closes it correctly.

The finding only applies to a hypothetical parent-driven programmatic open (`v-model:open="true"` set from outside), which isn't a usage pattern in this codebase. Since `open` defaults to `false` and existing callers are unaffected, no change needed. Happy to add the `openSelect`-observing sync later if we ever expose parent-driven opening.

gamalielhere (approved):

PR #5700, opened 2026-08-06
fix: re-translate perps Long/Short toggle on locale change (MEW-2112)

renerpdev:
## Summary

The perps **Long / Short** trade-direction toggle did not re-translate when the user switched the app language after the trade panel had mounted — the labels stayed in the initial locale (English by default), so a user on 中文 saw "Long/Short" instead of "做多/做空".

## Root cause

`orderSideButtons` in `usePerpsTradeForm.ts` was built as a **plain array**, so `t('perps.trade.long')` / `t('perps.trade.short')` were evaluated once at composable setup and frozen. vue-i18n's locale is reactive, but a plain array doesn't track it, so a later language switch never re-ran `t()`. (The zh/es translations already existed — this was never a missing-key issue.) Every other perps label uses `$t()` in-template or a `computed`, which is why only this toggle was affected.

## Fix

Make `orderSideButtons` a `computed`, so the labels re-evaluate on locale change like every other reactive label. The `v-for` consumer and the composable's return are unchanged (a computed ref auto-unwraps in the template).

## Testing

- `tests/unit/modules/perps/composables/usePerpsTradeForm.spec.ts` — added a reactivity test (a reactive-`t` spy asserts the toggle re-invokes `t()` after a locale flip); the existing key assertion switched to `unref`. **18/18 green.**
- `vue-tsc --noEmit -p tsconfig.app.json` — clean; eslint clean.
- Manual: perps trade panel, switch EN → 中文 / ES → Long/Short update live (做多/做空, Largo/Corto) without reload.

## Summary by CodeRabbit

- **Bug Fixes**
  - Long and Short trade labels now update immediately when the app language changes.

- **Tests**
  - Added coverage to verify that trade-side labels are correctly retranslated after switching locales.

### DIFF
--- a/changelog/fix-5700.md
+++ b/changelog/fix-5700.md
@@ -1,0 +1,1 @@
+fix: re-translate perps Long/Short toggle when the app language changes
--- a/changelog/fix-5700.md
+++ b/changelog/fix-5700.md
@@ -1,1 +1,0 @@
-fix: re-translate perps Long/Short toggle when the app language changes
--- a/src/modules/perps/composables/usePerpsTradeForm.ts
+++ b/src/modules/perps/composables/usePerpsTradeForm.ts
@@ -93,10 +93,13 @@ export function usePerpsTradeForm() {
 
   // ── State ──────────────────────────────────────────────────
 
-  const orderSideButtons: OrderSideButton[] = [
+  // `computed` (not a plain array) so the labels re-translate when the user
+  // switches language after the trade form has mounted. A plain array froze
+  // `t()` at setup and left Long/Short stuck in the initial locale (MEW-2112).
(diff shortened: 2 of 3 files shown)


changelog/fix-5700.md
```diff
@@ -0,0 +1 @@
+fix: re-translate perps Long/Short toggle when the app language changes
```
gamalielhere: no need for changelog

gamalielhere (approved):

changelog/fix-5700.md
renerpdev: Good catch — removed the changelog file in 0435fdfed8. Thanks!

PR #5701, opened 2026-08-06
fix: MEW-2111 prevent limit-order target price input overflow on manual entry

renerpdev:
## Summary

When a user manually types a value into the **Target price** input of a perps **Limit** order, the number could overflow past the right edge of its white rounded container instead of staying inside the box.

## Root cause

The Target price `<input>` in `ModulePerpsTrade.vue` sits in a flex row alongside the `$` prefix, but the input was `w-full` with no `min-w-0`. A flex item defaults to `min-width: auto` (≈ the input's intrinsic content width), so it could not shrink below that size — the element extended beyond its container and a long value rendered outside the box. The sibling reusable input (`PerpsAmount.vue`) does not have this bug because it already uses `min-w-0` on the input and `shrink-0` on the `$`.

## Fix

Match the working `PerpsAmount.vue` pattern on the Limit-order Target price input:
- add `min-w-0` to the `<input>` so it can shrink within the container and the long value scrolls internally
- add `shrink-0` to the `$` span so it is never squeezed

CSS/Tailwind-only change; no behavior/logic change.

## Testing

- `pnpm vue-tsc --noEmit -p tsconfig.app.json` → clean
- `eslint` on the changed file → clean
- Manual: Perps → Limit order → type a long number in Target price → value stays inside the box (scrolls internally), `$` fixed, no layout break.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the limit-order price field layout to keep the dollar prefix and input properly aligned.
  * Prevented the price input from shrinking or overflowing in constrained layouts.


changelog/fix-5701.md
```diff
@@ -0,0 +1 @@
+fix: prevent perps limit-order target price input from overflowing its container on manual entry
```
gamalielhere: we don't need to make change log files.

gamalielhere (changes requested):

changelog/fix-5701.md
renerpdev: Good catch, removed the changelog file in d346986334. Thanks!

gamalielhere (approved):

PR #5732, opened 2026-08-24
fix: [MEW-2185] guard private key access unlock against invalid input

renerpdev:
## What was wrong

On the Private Key access flow, the connect button is disabled for empty/invalid input, but the input's `@keyup.enter="unlock"` was **not** gated. Pressing **Enter** with an empty or invalid key (e.g. `0x1234`) ran `unlock()`, which built a wallet from wrong-length bytes and called the async `setWallet()` **fire-and-forget**. `setWallet` does `await newWallet.getAddress()` outside `unlock`'s try/catch, so `privateToAddress()` threw and the rejection escaped as an **unhandled promise rejection**:

```
Error: private key must be 32 bytes, hex or bigint, not object
```

Unhandled, 60 events since March, ongoing. Sentry: `APP-MEW-WEB-BY`.

## What changed

`src/modules/access/ModuleAccessPrivateKey.vue`:
- **Gate `unlock()` on the same `submitIsDisabled` condition as the button** — early-return so pressing Enter with empty/invalid input can't build a wallet from bad bytes.
- **`await setWallet(...)`** inside the existing try/catch, so any residual `getAddress()` / restriction failure surfaces as a handled toast instead of an unhandled throw.

Added `tests/unit/modules/access/ModuleAccessPrivateKey.spec.ts` — regression tests for the Enter path (invalid key blocked, empty input blocked, valid key connects).

## How to test

1. Open the app → access dialog → **Private Key** (ETH network).
2. Focus the input, press **Enter** with the field empty, then with an invalid key (`0x1234`) → no crash, nothing happens (button stays disabled).
3. Paste a valid private key, press **Enter** → connects normally.

Unit: `pnpm vitest run tests/unit/modules/access/ModuleAccessPrivateKey.spec.ts` (3/3). `vue-tsc` + eslint clean.

## Sentry

https://myetherwallet-inc.sentry.io/issues/APP-MEW-WEB-BY

## Summary by CodeRabbit

* **New Features**
  * Pressing Enter in access password and private-key fields now submits when the input is valid and submission is enabled.
  * Submission is automatically disabled when required credentials are missing or invalid.

* **Bug Fixes**
  * Improved private-key unlock handling so wallet setup errors are surfaced through existing notifications and error reporting.

* **Tests**
  * Added coverage for Enter-key submission, disabled states, credential validation, and successful private-key unlocks.

### DIFF
--- a/src/modules/access/ModuleAccessPrivateKey.vue
+++ b/src/modules/access/ModuleAccessPrivateKey.vue
@@ -15,7 +15,8 @@
             :aria-label="
               $t('access_wallet_private_key.private_key_input_label')
             "
-            @keyup.enter="unlock"
+            @enter="unlock"
+            :submit-disabled="submitIsDisabled"
             :error-message="errorMessages"
           />
           <div class="flex align-center justify-center">
@@ -119,10 +120,6 @@ const isValidPrivateKey = computed<boolean>(() => {
 })
 
 const unlock = async () => {
-  // The connect button is disabled for empty/invalid input, but the input's
-  // `@keyup.enter` is not — bail on the same condition so pressing Enter with an
-  // invalid key can't build a wallet from bad bytes and crash (MEW-2185).
-  if (submitIsDisabled.value) return
   // TODO: remove hardcoded network id
   let wallet
   try {
--- a/src/modules/access/ModuleAccessPrivateKey.vue
+++ b/src/modules/access/ModuleAccessPrivateKey.vue
@@ -118,7 +118,11 @@ const isValidPrivateKey = computed<boolean>(() => {
   }
 })
 
-const unlock = () => {
+const unlock = async () => {
+  // The connect button is disabled for empty/invalid input, but the input's
+  // `@keyup.enter` is not — bail on the same condition so pressing Enter with an
+  // invalid key can't build a wallet from bad bytes and crash (MEW-2185).
+  if (submitIsDisabled.value) return
   // TODO: remove hardcoded network id
   let wallet
   try {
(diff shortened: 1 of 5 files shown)


src/modules/access/ModuleAccessPrivateKey.vue
```diff
@@ -118,7 +118,11 @@ const isValidPrivateKey = computed<boolean>(() => {
   }
 })
 
-const unlock = () => {
+const unlock = async () => {
+  // The connect button is disabled for empty/invalid input, but the input's
+  // `@keyup.enter` is not — bail on the same condition so pressing Enter with an
+  // invalid key can't build a wallet from bad bytes and crash (MEW-2185).
+  if (submitIsDisabled.value) return
```
gamalielhere: it may be better to disable it on the component itself, not on individual modules that uses the component.

gamalielhere (changes requested):

src/modules/access/ModuleAccessPrivateKey.vue
renerpdev: Moved the gate into `AppInput`. It now exposes an `@enter` event that is withheld when `:submit-disabled` is true, so modules gate Enter on the same condition as their submit button without re-implementing the guard. `ModuleAccessPrivateKey` and `ModuleAccessKeystore` both adopt `@enter` + `:submit-disabled` (keystore had the same ungated-Enter path). Per-module guard in `unlock()` removed. Added an `AppInput` spec for the gate; module spec now asserts the wiring. Commit 1ef82a1a58.

gamalielhere (approved):

